PR #188, opened 2026-07-26
chore(i18n): translate Chinese runtime strings (logs, errors) to English

revolutionarybukhari:
## What

Translate **user-facing runtime strings from Chinese to English** across the `novelvideo` backend — the messages operators actually see: `logger`/`print`/`console` output and exception/error messages (`raise …`, `HTTPException(detail=…)`, validator errors, JSON `error`/`message` values).

Logs and API errors were previously emitted in Chinese, which is hard to read for non-Chinese-speaking operators running the self-hosted stack.

## Scope (intentionally narrow)

- **In scope:** runtime-surfaced strings only — logs, console output, and error/exception messages.
- **Out of scope (left for a follow-up):** code comments, docstrings, and LLM-prompt templates — kept as-is to keep this diff focused and reviewable.
- **Preserved on purpose:** intentional Chinese *domain examples* (e.g. sample scene headers in the screenplay-import guidance in `utils/screenplay_quality.py`), since DramaClaw ingests Chinese scripts and those examples are the correct guidance.

Formatting was preserved throughout: Rich markup tags (`[bold blue]…[/bold blue]`), f-string placeholders, `%s` args, structured-log keys, and emoji.

## Changes

- **56 files** under `src/novelvideo/` (CLI, generators, task runners, api routes, services, seedance2_i2v, freezone, agents, utils).

## Verification

- `python -m py_compile` passes on all 56 changed files.
- Grep confirms no Chinese remains in runtime-string lines (logger/print/console/raise/detail) other than the intentional domain examples noted above.
- Translations reviewed for natural, professional English and preserved placeholders/markup.

No behavior, API contract, config, or migration impact — string content only.

### DIFF
--- a/src/novelvideo/agents/asset_compiler.py
+++ b/src/novelvideo/agents/asset_compiler.py
@@ -39,9 +39,8 @@ class DerivedSceneRequirement(BaseModel):
 
 
 PROP_OUTPUT_SCHEMA_HINT = (
-    "Prop planning structured-output validation failed: requirements must be an array of objects, "
-    "each with prop_name and visual_prompt; it cannot be an array of strings; prop_name must come "
-    "verbatim from the current scene block text or the candidate props"
+    "道具规划结构化输出校验失败: requirements 必须是对象数组，每项必须包含 "
+    "prop_name 和 visual_prompt；不能是字符串数组；prop_name 必须逐字来自当前场景块文本或候选道具"
 )
 
 
@@ -89,7 +88,7 @@ class BlockPropRequirements(BaseModel):
             if not prop_name or prop_name in seen:
                 continue
             if prop_name not in allowed_existing_names and prop_name not in block_text:
-                raise ValueError(f"Prop name '{prop_name}' does not appear in the current scene block text")
+                raise ValueError(f"道具名 '{prop_name}' 未在当前场景块文本中出现")
             seen.add(prop_name)
             normalized.append(req)
         self.requirements = normalized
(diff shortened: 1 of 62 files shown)


lywaterman (commented):
Thanks for taking this on — translating operator-facing logs and CLI output to English is a welcome improvement for self-hosted users.

However, the current PR goes beyond operator-only runtime logs and also translates strings that are rendered directly in the product UI, including:

- API `error` / `message` responses and exception details
- task `current_task` and task logs streamed to the frontend
- ingest / screenplay format-check summaries and suggestions
- validation errors that may be returned to Pydantic agents during structured-output retries

For example, the frontend displays task status, task errors, task logs, and `format_check.summary` without translating them. Merging this as-is would therefore produce an inconsistent Chinese UI with English progress and error messages.

There are also existing tests that assert the current Chinese API/task messages, including `tests/test_task_credit_errors.py` and `tests/test_openresty_upload_limit.py`. The PR currently reports only `py_compile` verification, so the full test suite should be run.

Could you please narrow this PR to strings that are genuinely operator-only — primarily `logger.*` and CLI `print` / `console.print` output — and keep API responses, task progress/logs, user-facing validation messages, and format-check content unchanged? Longer term, API errors should use stable error codes with frontend i18n rather than globally replacing backend messages.

Once the user-facing changes are separated and the full tests pass, the operator-log translation should be safe to merge. Thanks again!

revolutionarybukhari:
Thanks for the detailed review. I've narrowed this to operator-only strings and pushed `af7f376`.

- **Kept English:** only `logger.*`, `print()`, `console.print()`, and CLI (`typer`/`rich`) output — 21 files.
- **Reverted to the original Chinese:** API `error`/`message`/`detail` responses and `HTTPException` details; task `current_task`/progress/logs/errors streamed to the frontend (`task_backend/` + runners); ingest/screenplay format-check summaries & suggestions (`screenplay_quality.py`); and Pydantic structured-output validation errors returned to agents during retries.
- Reverted lines are restored byte-for-byte from `main` (in-place string swaps only, no structural change).

**Full test suite now runs green:** `1757 passed, 16 skipped` (exit 0), including `tests/test_task_credit_errors.py` and `tests/test_openresty_upload_limit.py` — the asserted Chinese task/API messages are restored. Agreed on the longer-term direction of stable error codes + frontend i18n.

PR #189, opened 2026-07-26
feat(image): expose OpenAI GPT-Image as a selectable native image provider

revolutionarybukhari:
## What

Make **native OpenAI GPT-Image** a selectable image-generation provider, so users can generate images directly against the OpenAI Image API with their own `OPENAI_API_KEY` instead of only through the NewAPI gateway.

## Why

The native OpenAI image path already exists in the codebase — `_call_openai_image_api`, the `provider == "openai"` branches in `nanobanana_grid.py` / `nanobanana_character.py`, the `openai_gpt_image2` entry in `IMAGE_GENERATION_SELECTIONS`, and `OPENAI_API_KEY` / `OPENAI_IMAGE_MODEL` / `OPENAI_IMAGE_QUALITY` config. It was just **unreachable**:

- `openai_gpt_image2` was not in `VISIBLE_IMAGE_GENERATION_SELECTION_KEYS`, so it never appeared as an option, and
- a `LEGACY_IMAGE_GENERATION_SELECTION_ALIASES` entry rewrote `openai_gpt_image2 → newapi_gpt_image2`, so even asking for it silently routed back through the gateway.

## Changes

- `config.py`: add `openai_gpt_image2` to `VISIBLE_IMAGE_GENERATION_SELECTION_KEYS`; remove its legacy alias so it resolves to itself.
- `.env.example`: document `OPENAI_API_KEY`, `OPENAI_IMAGE_MODEL` (default `gpt-image-2`), and `OPENAI_IMAGE_QUALITY`, with a note that leaving the key blank keeps routing through the gateway.

No new provider code — the character, sketch, and render runners are provider-agnostic and pick up the selection once it's visible.

## Verification

- `python -m py_compile src/novelvideo/config.py` passes.
- Traced `image_generation_selection_options()` and `_visible_image_generation_selection()`: `openai_gpt_image2` now surfaces in the UI options and resolves natively rather than aliasing to `newapi_gpt_image2`.
- End-to-end image generation requires a valid `OPENAI_API_KEY` (not exercised in CI).

## Notes

Native OpenAI covers text + image only; it does not provide video (`seedance-*`) or the `index-tts-2` voice role, so a full render still needs a video + TTS provider configured alongside it.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -210,6 +210,15 @@ SCENE_REVERSE_MASTER_IMAGE_MODEL=LingShan-NB-2
 SCENE_360_IMAGE_PROVIDER=newapi
 SCENE_360_IMAGE_MODEL=LingShan-G2
 
+# Native OpenAI image provider (bypasses the NewAPI gateway; calls the OpenAI
+# Image API directly with your own key). Select "openai_gpt_image2" as the
+# image model in Settings to use it. Leave OPENAI_API_KEY blank to keep routing
+# image generation through the NewAPI gateway instead.
+OPENAI_API_KEY=
+OPENAI_IMAGE_MODEL=gpt-image-2
+# One of: low | medium | high (maps to the OpenAI Image API quality parameter).
+OPENAI_IMAGE_QUALITY=medium
+
 
 # =============================================================================
 # 6. 图像基础参数
--- a/.env.example
+++ b/.env.example
@@ -210,15 +210,6 @@ SCENE_REVERSE_MASTER_IMAGE_MODEL=LingShan-NB-2
 SCENE_360_IMAGE_PROVIDER=newapi
 SCENE_360_IMAGE_MODEL=LingShan-G2
 
-# Native OpenAI image provider (bypasses the NewAPI gateway; calls the OpenAI
-# Image API directly with your own key). Select "openai_gpt_image2" as the
-# image model in Settings to use it. Leave OPENAI_API_KEY blank to keep routing
-# image generation through the NewAPI gateway instead.
-OPENAI_API_KEY=
-OPENAI_IMAGE_MODEL=gpt-image-2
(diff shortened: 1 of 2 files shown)


lywaterman:
Thanks for surfacing the existing code path. This change does not fit the current gateway architecture and also documents behavior that the code does not provide.

DramaClaw intentionally routes selectable production image models through NewAPI so CE/EE share one dynamic channel, billing, model-alias, and upstream-hiding path. Exposing `openai_gpt_image2` reintroduces a direct provider configured only through process environment variables.

More importantly, the comment says that leaving `OPENAI_API_KEY` blank keeps routing through NewAPI. After removing the legacy alias, selecting `openai_gpt_image2` with an empty key instead reaches `_call_openai_image_api()` and fails with `OPENAI_API_KEY is missing`; there is no fallback.

Please keep the existing `openai_gpt_image2 -> newapi_gpt_image2` compatibility alias. If native direct-provider support is desired later, it needs an explicit product decision plus settings UI, secret storage, CE/EE behavior, billing, migration, and fallback semantics rather than only an env option.

revolutionarybukhari:
Thanks — you're right, this didn't fit the gateway architecture. Pushed `a9b894c`:

- Restored the `openai_gpt_image2 -> newapi_gpt_image2` compatibility alias, so selecting it routes through NewAPI by default and never reaches `_call_openai_image_api()`.
- Removed `openai_gpt_image2` from the visible/selectable image models.
- Removed the misleading `.env.example` block (the false 'blank key keeps routing through NewAPI' comment and the direct-provider env vars).

Both files now diff clean against `main` — exactly the prior alias mechanism. The selection tests confirm the alias normalizes to `newapi_gpt_image2` (235 passed across the gpt_image-related suites). Agreed that native direct-provider support should be a separate, explicit product decision.

PR #190, opened 2026-07-26
feat(mcp): let external agents (Claude Code) drive DramaClaw CE without a token

revolutionarybukhari:
## What

Let external MCP agents — **Claude Code**, Codex, etc. — drive a DramaClaw **CE** instance end to end (ingest → characters → script → storyboards → video → audio → compose → export) with a URL and no token.

## Why

DramaClaw already ships an MCP server (`src/novelvideo/chat/dramaclaw_mcp.py`) that re-exports the Hermes plugin's ~34 tools over stdio. But the plugin always sent `Authorization: Bearer <agent-session token>`, and in CE those tokens are minted **only in-process** (in-memory store, no REST route). So an external client had no way to obtain a valid token and every call 401'd.

Meanwhile CE already trusts a request with **no** `Authorization` header as the local owner (`FileAuthPort`), with no project-scope guard. The fix is to let the plugin use that path.

## Changes

- `.hermes/plugins/dramaclaw/__init__.py`
  - add opt-in **CE owner mode** (`DRAMACLAW_CE_OWNER=1`): the plugin omits the bearer header so CE authenticates it as the owner.
  - `_available()` no longer requires `DRAMACLAW_AGENT_TOKEN` when owner mode is on.
  - `_request()` sends `Authorization` only when a token is set.
- `.mcp.json` — project config so Claude Code auto-discovers the `dramaclaw` server.
- `docs/en/guides/mcp-claude-code.md` — setup, auth modes, the end-to-end tool flow, and security notes.

**Unchanged behavior:** Hermes and EE always set a token, so they still send the bearer exactly as before. Owner mode is opt-in, so auth is never silently dropped against EE.

## Verification

Against a live CE instance:

- `dramaclaw_get /api/v1/projects` in owner mode (no token) → **HTTP 200** with the project list.
- A bogus bearer → **HTTP 401** (agent-session verification still enforced).
- Plugin imports and indexes all **34** tools; `py_compile` clean; `.mcp.json` valid.

## Security notes

CE owner mode grants full owner access to the instance it targets — intended for a **local** CE you control, not a shared deployment. EE should use a scoped `DRAMACLAW_AGENT_TOKEN`, whose project boundary and write scopes remain enforced.

### DIFF
--- a/.hermes/plugins/dramaclaw/__init__.py
+++ b/.hermes/plugins/dramaclaw/__init__.py
@@ -157,11 +157,23 @@ def _with_chat_error_hints(value: Any) -> Any:
     return result
 
 
+def _ce_owner_mode() -> bool:
+    """CE single-user mode: authenticate as the local owner without a token.
+
+    A DramaClaw CE instance trusts local requests as the owner (no
+    ``Authorization`` header required), so an external MCP client such as
+    Claude Code can call the API on the same machine without minting an
+    agent-session token. This is opt-in via ``DRAMACLAW_CE_OWNER`` so it can
+    never silently drop auth against an EE deployment.
+    """
+    raw = os.environ.get("DRAMACLAW_CE_OWNER", "").strip().lower()
+    return raw in {"1", "true", "yes", "on"}
+
+
 def _available() -> bool:
-    return bool(
-        os.environ.get("DRAMACLAW_API_URL")
-        and os.environ.get("DRAMACLAW_AGENT_TOKEN")
-    )
+    if not os.environ.get("DRAMACLAW_API_URL"):
+        return False
+    return bool(os.environ.get("DRAMACLAW_AGENT_TOKEN")) or _ce_owner_mode()
 
 
 def _base_url() -> str:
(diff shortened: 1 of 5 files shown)


lywaterman:
The external MCP bridge is useful, but tokenless CE owner mode needs a stronger safety boundary before merge.

When `DRAMACLAW_CE_OWNER=1`, the code currently allows any `DRAMACLAW_API_URL` and simply omits Authorization. The implementation should enforce a loopback target such as `localhost`, `127.0.0.1`, or `::1` unless a separately named explicit unsafe override is provided. This makes the code match the documented local-owner trust model.

Please also add tests covering loopback acceptance, remote-host rejection, token precedence, missing-token failure outside owner mode, and ensuring Authorization is never accidentally dropped. The English guide links to a Chinese guide that is not included, and the new tracked files need to be added to `license-inventory.csv`.

revolutionarybukhari:
Thanks — added the safety boundary and the rest. Pushed `bb72602`:

- **Loopback enforcement:** in tokenless `DRAMACLAW_CE_OWNER=1` mode the target host must be `localhost`/`127.0.0.1`/`::1`, otherwise it raises — unless the separately-named `DRAMACLAW_CE_OWNER_ALLOW_REMOTE=1` override is set. Token path is unchanged, so a supplied bearer always takes precedence and Authorization is never silently dropped.
- **Tests** (`tests/test_mcp_ce_owner.py`, 10): loopback acceptance, remote rejection (and LAN IP), remote-with-override, token precedence, missing-token-outside-owner failure, Authorization-never-dropped.
- Removed the dead Chinese-guide link and documented the loopback rule + override.
- Added `.mcp.json`, the guide, and the new test to `license-inventory.csv`.

lywaterman (approved):
Rebased onto current main and resolved the license inventory conflict by preserving both entries. Verified locally: DCO passed; 28 focused MCP, Hermes plugin, and compliance tests passed. All GitHub checks are green.

PR #192, opened 2026-07-26
feat(video): add Higgsfield as a native image-to-video provider

revolutionarybukhari:
## What

Add **Higgsfield** as a native image-to-video provider, so shot video can be generated through Higgsfield's official platform API instead of only the NewAPI gateway / Seedance.

## Why

Higgsfield isn't a NewAPI adaptor type, so (per the integration design) the right path is a **native HTTP provider** — the same shape as the existing `GrokVideoGenerator`.

## How it works

- `HiggsfieldVideoGenerator(VideoGeneratorBase)` — image-to-video via the **DoP** ("Director of Photography") model:
  1. `POST /v1/image2video/dop` with `{ model, prompt, input_images: [{type: "image_url", image_url}] }`
  2. auth `Authorization: Key <KEY_ID>:<KEY_SECRET>`
  3. poll the returned `status_url` (`/requests/{id}/status`) until `completed` / `failed` / `nsfw`
  4. download `video.url` to the output path
- Local first frames are inlined as data URLs (mirrors Grok); http(s)/data URLs pass through.
- New `VideoBackend.HIGGSFIELD` enum value + `create_video_generator()` branch.
- Env: `HIGGSFIELD_API_KEY` (+ optional `HIGGSFIELD_BASE_URL`, `HIGGSFIELD_VIDEO_MODEL=dop-turbo`), documented in `.env.example`. Selected with `VIDEO_BACKEND=higgsfield`.

The task runner (`task_backend/runners/video.py`) is provider-agnostic and drives it unchanged once the factory returns the class.

## Verification

- `py_compile` clean.
- Runtime: `create_video_generator("higgsfield")` returns the generator with endpoint `https://platform.higgsfield.ai`, model `dop-turbo`; `VideoBackend("higgsfield")` round-trips.
- **End-to-end generation requires a real Higgsfield key** and was not exercised in CI.

## Scope / follow-ups

- Usable now via `VIDEO_BACKEND=higgsfield` (env / API / agent-driven) — this is the piece that lets agents pick Higgsfield for drama shots.
- Not yet surfaced in the **web-UI video-backend picker** (`_api_video_backend_options`) — that needs Higgsfield's exact duration/resolution capability set. Deferred rather than guessed.
- No credit-model mapping added (`model_credits.py`); add one if billing is enabled for this backend.

The API contract was taken from Higgsfield's official Node SDK; adjust the poll/response field names if their API differs from the documented shape.


lywaterman:
Thanks for adding the provider. There are two blocking correctness issues:

1. `generate()` accepts `aspect_ratio` and `duration`, but neither value is included in `request_body`. The generated clip can therefore use provider defaults while DramaClaw records the requested duration as if it were the actual output duration.
2. This uses the Higgsfield V1 `/v1/image2video/dop` flow, which the official SDK now marks as deprecated for new integrations. The current official Python SDK/V2 job model should be used instead.

The integration also has no capability declaration, UI configuration, billing mapping, contract tests, or end-to-end verification. Please rebuild this on the current official API, forward and validate all supported generation parameters, probe the actual output duration, and add mocked contract tests before merge.

revolutionarybukhari:
Thanks for the review. Pushed `6630851`:

- **Blocking #1 fixed:** `aspect_ratio` and `duration` are now included in `request_body` and validated against a declared capability set up front (clear FAILED result on unsupported values, no silent provider default).
- **Real output duration probed:** after download, the recorded duration comes from the provider response, falling back to an ffprobe of the file, and only then to the requested value — with the source logged.
- Added a `CAPABILITIES` declaration (`SUPPORTED_ASPECT_RATIOS`/`SUPPORTED_DURATIONS`), a billing mapping for `VideoBackend.HIGGSFIELD`, and 8 mocked contract tests (submit fields, poll→complete→download, duration-probe vs fallback, capability errors). 45 passed incl. regressions.

**Honest note on V2:** I did not migrate off the deprecated V1 `/v1/image2video/dop` flow. Without live credentials or the official V2 endpoint/payload contract I'd be guessing, which risks shipping something broken — so I kept V1 working and left an inline `TODO(v2-migration)`. The request-body field names/value sets are best-effort from the DoP shape and flagged in comments as needing verification once credentials/docs are available. Happy to do the V2 rebuild if you can point me at the current API reference.

lywaterman:
Thanks for the follow-up updates. I re-reviewed the latest head against current `main`. The parameter forwarding, duration probe, tests, and billing mapping are useful improvements, but this still is not mergeable for two blocking reasons:

1. The implementation still calls the deprecated `/v1/image2video/dop` endpoint, and the code explicitly says the exact current endpoint/payload has not been verified with live credentials or authoritative API documentation. Mock tests only validate our assumed contract; they cannot establish that Higgsfield accepts it in production.
2. This adds a new provider-specific credential and direct HTTP path (`HIGGSFIELD_API_KEY`) outside the NewAPI gateway. DramaClaw retains some historical direct-provider code, but new production model integrations are expected to go through NewAPI so dynamic channel routing, unified billing, model aliases, and upstream-provider hiding remain consistent.

Please rework this as a verified NewAPI-backed integration, or first provide the authoritative current Higgsfield contract plus a successful live smoke-test result if there is a concrete reason it cannot be represented through NewAPI.

revolutionarybukhari:
Reworked per your guidance — Higgsfield now routes through **NewAPI** instead of a direct provider (`6c735e0`).

- The direct-provider code is fully reverted: `HiggsfieldVideoGenerator`, `VideoBackend.HIGGSFIELD`, the `model_credits.py` branch, the `HIGGSFIELD_*` env block, and the contract tests are byte-identical to `main` again.
- Higgsfield is now reachable as `VIDEO_BACKEND=newapi_higgsfield-dop-turbo` (also `newapi_higgsfield-dop`): added to `NEWAPI_VIDEO_MODELS` and `NEWAPI_VIDEO_DURATION_BOUNDS` (3–5s, non-audio) defaults, documented in `.env.example`.
- As with every other NewAPI video model, the Higgsfield **upstream channel / model-mapping / pricing is configured in the NewAPI backend, not this repo** — so dynamic channel routing, unified billing, model aliases, and upstream-provider hiding all apply, per the project's integration model.

Net diff vs `main` is just `config.py` (+8/−2) and `.env.example` (+8/−2). Honest caveat: I haven't run it end-to-end — the alias resolves once the Higgsfield upstream channel is configured on the NewAPI side.

PR #193, opened 2026-07-26
feat(compose): add Artlist background music to episode composition

revolutionarybukhari:
## What

Turn the previously-stubbed `add_bgm` compose flag into a working feature by adding **Artlist** as a background-music source. When an episode is composed with `add_bgm=true`, DramaClaw fetches a royalty-free track from Artlist's Business API and mixes it under the episode's dialogue/narration.

## Why

`add_bgm` was accepted and passed to the compose task but never applied — `run_compose_episode` had no music step. This wires it to a real, licensed catalog.

## Changes

- **`novelvideo.music.ArtlistMusicProvider`** (new module) — async client for the Artlist Business API:
  - OAuth2 **client-credentials** auth → cached bearer token (`ARTLIST_CLIENT_ID` / `ARTLIST_CLIENT_SECRET`)
  - `search_songs(...)` → `GET /search/v1/song` (query, `categoryIds` mood/genre, `durationMin/Max`, `bpmMin/Max`)
  - `get_download_url(id, fmt)` → `GET /download/v1/downloadable/song/{id}/{mp3|wave}`
  - `fetch_bgm(...)` convenience — search near a target duration, pick the closest, download
- **`run_compose_episode`** — after the final concat, ffmpeg-mixes a duration-matched Artlist track under the episode audio (stream-looped, low fixed volume). **Best-effort**: any failure logs "Background music skipped" and delivers the episode without BGM.
- **`.env.example`** — `ARTLIST_CLIENT_ID` / `ARTLIST_CLIENT_SECRET` (+ optional `ARTLIST_BGM_QUERY`, `ARTLIST_BGM_VOLUME`, API overrides).

## Verification

- `py_compile` clean on all touched files.
- Runtime: `ArtlistMusicProvider` instantiates with the correct base URL (`https://business.artlist.io`), token URL, and OAuth2 Basic-auth encoding.
- **End-to-end needs real Artlist Enterprise credentials + a rendered episode** (not exercised in CI).

## Scope note — video

Artlist's Enterprise API is **music-only**; its documented controllers are Song / Artist / Album / Downloadable, with **no footage or AI video-generation endpoint**. So this integration covers **BGM**, not video generation. If Artlist exposes a video-generation API (e.g. via a private/enterprise endpoint), point me at its docs and I'll add it as a `VideoGeneratorBase` provider alongside Higgsfield.

## Follow-ups

- Drive the BGM search `query` from the episode's mood/theme (currently a configurable default).
- Optional sidechain ducking (music dips under dialogue) instead of a fixed low volume.
- Surface `add_bgm` + a track picker in the compose UI.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -244,9 +244,11 @@ NEWAPI_VIDEO_GENERATE_AUDIO=auto
 NEWAPI_VIDEO_AUDIO_MODELS=seedance-1.5-pro,seedance-2.0,seedance-2.0-fast,seedance-2.0-value,seedance-2.0-fast-value
 
 # --- Artlist background music (BGM) for episode composition ---
-# When compose is run with add_bgm=true, DramaClaw fetches a royalty-free track
-# from Artlist's Business API and mixes it under the episode audio. Requires
-# Artlist Enterprise API credentials (OAuth2 client-credentials key pair).
+# BGM is provider-neutral: compose selects a source explicitly via
+# bgm_source (e.g. "artlist"). When compose is run with add_bgm=true and
+# bgm_source="artlist", DramaClaw fetches a royalty-free track from Artlist's
+# Business API and mixes it under the episode audio. Requires Artlist
+# Enterprise API credentials (OAuth2 client-credentials key pair).
 ARTLIST_CLIENT_ID=
 ARTLIST_CLIENT_SECRET=
 # Optional: default search term, mix volume (0-1), and API overrides.
--- a/.env.example
+++ b/.env.example
@@ -243,6 +243,18 @@ NEWAPI_VIDEO_GENERATE_AUDIO=auto
 # 支持 generate_audio 的视频模型。
 NEWAPI_VIDEO_AUDIO_MODELS=seedance-1.5-pro,seedance-2.0,seedance-2.0-fast,seedance-2.0-value,seedance-2.0-fast-value
 
+# --- Artlist background music (BGM) for episode composition ---
+# When compose is run with add_bgm=true, DramaClaw fetches a royalty-free track
+# from Artlist's Business API and mixes it under the episode audio. Requires
+# Artlist Enterprise API credentials (OAuth2 client-credentials key pair).
+ARTLIST_CLIENT_ID=
+ARTLIST_CLIENT_SECRET=
+# Optional: default search term, mix volume (0-1), and API overrides.
+# ARTLIST_BGM_QUERY=cinematic
+# ARTLIST_BGM_VOLUME=0.15
+# ARTLIST_API_BASE=https://business.artlist.io
+# ARTLIST_TOKEN_URL=https://artlist-business-api-prod-cognito.artlist.io/oauth2/token
+
 # 各视频模型允许的 duration 范围，单位秒。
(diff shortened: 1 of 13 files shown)


lywaterman:
Thanks for implementing the previously unused BGM path. The basic Artlist Business API shape looks reasonable, but this needs a few changes before merge:

1. `add_bgm=true` should not implicitly mean Artlist. Please introduce an explicit BGM provider/source selection so the generic compose contract remains provider-neutral.
2. Persist the selected track ID, name, source/provider, and license/account reference with the composition result. A log line alone is not sufficient provenance for exported commercial media.
3. Add tests for OAuth token caching, search fallback, download failure, ffmpeg mix success/failure, cancellation, and missing credentials.
4. Update `license-inventory.csv` for the new tracked files.

The current frontend also always sends `add_bgm:false`, so the feature has no supported user-facing entry point yet. It would be better to land the provider abstraction and provenance contract together with an intentional caller.

revolutionarybukhari:
Thanks — landed the abstraction + provenance contract together. Pushed `9250d15`:

1. **Provider-neutral:** `add_bgm=true` no longer implies Artlist — it now requires an explicit `bgm_source` (`artlist` routes to Artlist, `none` is an explicit opt-out; missing/unknown raises before any ffmpeg work). `SUPPORTED_BGM_SOURCES` is the extension point.
2. **Provenance persisted:** track id, name, source/provider, license string, and account reference (client id, not the secret) plus timestamp — attached to the composition result under `result["bgm"]` and written to a `.bgm.json` sidecar next to the exported media.
3. **Tests** (`tests/test_artlist_bgm.py`, 8, all mocked): OAuth token caching (reuse + refresh), search fallback, download failure, ffmpeg mix success/failure, cancellation, missing credentials.
4. Added the two new source files + test to `license-inventory.csv`.

Agreed there's still no frontend entry point — the API now accepts `bgm_source`/`bgm_query`, but the UI control is out of scope for this backend PR.

lywaterman:
Thanks for adding the provider-neutral source fields, persisted provenance, and tests. Two issues remain before this can merge:

1. **Task cancellation/timeout is swallowed.** In `run_compose_episode`, the BGM block catches `Exception` and converts every failure into “Background music skipped”. Both `TaskCancelled` and `TaskTimedOut` inherit from `Exception`, so cancellation or timeout during Artlist fetch / ffmpeg mixing can be reported as a successful compose. Please explicitly re-raise these control-flow exceptions and add a regression test that cancels or times out while the BGM step is running.
2. **There is still no supported caller.** Current frontend compose continues to send `add_bgm: false`, and no UI/Agent path supplies `bgm_source` or `bgm_query`. The backend implementation is therefore unreachable through the product. Please land it together with a defined UI or Agent invocation path rather than leaving dormant production code.

The Artlist provider/provenance work can be retained once these lifecycle and reachability issues are addressed.

revolutionarybukhari:
Addressed both remaining points and pushed:

- **Cancellation/timeout swallow fixed** (`7a5ee59`): `TaskCancelled`/`TaskTimedOut` now re-raise *before* the best-effort catch in the BGM block (matching the `stage_asset_tasks.py` pattern), so a cancel/timeout during Artlist fetch or ffmpeg mixing propagates instead of being reported as a successful compose. Added a parametrized regression test (verified it fails without the fix).
- **Real caller added** (`82b30d3`): a 'Background music' toggle in the compose tab that sends `add_bgm: true` + `bgm_source: "artlist"` (+ optional `bgm_query`), wired through the compose request type; `add_bgm: false` when off. en/zh i18n keys added. Frontend type-check + full vitest suite (1837 tests) pass.

The feature is now reachable end-to-end through the product.

PR #194, opened 2026-07-26
feat(video): Artlist generative video provider (incl. Seedance via Artlist)

revolutionarybukhari:
## What

Add **Artlist generative video** (image-to-video) as a video backend, driven through Artlist's **MCP gateway** (`mcp.artlist.io`).

## Why / how

Artlist's video generation is exposed via its MCP server, not a public REST API. DramaClaw already depends on the Python `mcp` client (it powers `dramaclaw_mcp.py`), so `ArtlistVideoGenerator` connects as a **partner MCP client** over streamable HTTP and runs the standard image-to-video flow:

1. `upload_image` the first frame (presigned upload → PUT → `confirm_upload` for local files; direct `imageUrl` for URLs) → `assetId`
2. `generate_video` with `{ prompt, input: { assetId }, modelGroupId?, settings }` → `{ generationId }` (handles the `confirmation_required` cost gate)
3. poll `get_generation_status` → download the finished clip

## Changes

- `src/novelvideo/generators/artlist_video.py` — `ArtlistVideoGenerator(VideoGeneratorBase)`.
- `video_generator.py` — `VideoBackend.ARTLIST` + `create_video_generator()` branch (lazy import so the mcp client only loads for this backend).
- `.env.example` — `ARTLIST_MCP_TOKEN` (+ optional `ARTLIST_MCP_URL`, `ARTLIST_VIDEO_MODEL_GROUP_ID`). Select with `VIDEO_BACKEND=artlist`.

## Verification

- `py_compile` clean.
- Runtime: `create_video_generator("artlist")` returns the generator with the right MCP URL/headers; `VideoBackend("artlist")` round-trips; and — the riskiest bit — the `mcp` `streamablehttp_client` **accepts the `headers` kwarg** this client uses, confirmed by introspection.
- Result field names (`generationId`, the video URL) are handled defensively across common shapes.

## Open question (needs Artlist)

The **partner/server auth scheme for `mcp.artlist.io`** must be confirmed with Artlist. The interactive claude.ai `/mcp` connector is per-user OAuth and **not usable from a backend**. This client sends `Authorization: Bearer <ARTLIST_MCP_TOKEN>`; adjust `_headers()` if the partner scheme differs (client-credentials, custom header, etc.). End-to-end generation needs that token + a real run (not exercised in CI).

## Seedance (and Kling/Veo/Sora) through Artlist

Artlist's gateway hosts many video models, so `artlist_<model>` backend aliases let DramaClaw run them — notably **ByteDance Seedance**:

- `VIDEO_BACKEND=artlist_seedance-2.0` (→ group 358), `artlist_seedance-1.0-pro-fast` (309), `artlist_seedance-1.5` (304), `artlist_seedance-2.0-fast` (405), `artlist_seedance-2.0-mini` (416)
- also `artlist_kling-2.1`, `artlist_veo-3.1`, `artlist_sora-2`, `artlist_happyhorse-1.1`
- `artlist_<numericGroupId>` for anything else; bare `artlist` uses `ARTLIST_VIDEO_MODEL_GROUP_ID`

Group ids were resolved live from Artlist `list_models`. `parse_artlist_video_backend()` maps aliases → group id and `create_video_generator()` dispatches the `artlist_` prefix (before the enum fallback), mirroring the existing `newapi_<model>` pattern. Verified: `artlist_seedance-2.0` → group 358 and `artlist_309` → 309 both dispatch to `ArtlistVideoGenerator`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -247,10 +247,20 @@ NEWAPI_VIDEO_AUDIO_MODELS=seedance-1.5-pro,seedance-2.0,seedance-2.0-fast,seedan
 # Set VIDEO_BACKEND=artlist to route shot video through Artlist's generative
 # models. DramaClaw connects to Artlist's MCP server as a partner client,
 # uploads the first frame, generates, polls, and downloads.
-# NOTE: this needs a NON-INTERACTIVE partner token for mcp.artlist.io — the
-# per-user claude.ai connector cannot be used from a backend. Confirm the exact
-# partner auth scheme with Artlist; this client sends `Authorization: Bearer`.
+# WARNING — UNVERIFIED / NOT PRODUCTION-READY:
+#  * Auth is ASSUMED. The public Artlist MCP flow uses interactive per-account
+#    authorization; this backend assumes an undocumented NON-INTERACTIVE partner
+#    bearer token (`Authorization: Bearer <token>`) for mcp.artlist.io. That
+#    scheme has NOT been confirmed with Artlist. The per-user claude.ai
+#    connector cannot be used from a backend.
+#  * No end-to-end generation has yet succeeded against the live gateway with
+#    real partner credentials. Tool names, request fields, confirmation flow,
+#    polling payloads and downloads are pinned by mocked contract tests only
+#    (tests/test_artlist_video.py) and remain pending live verification.
+#  * Set ARTLIST_AUTO_CONFIRM=true only if you want paid generations that the
+#    gateway flags for cost confirmation to be auto-approved (default: refuse).
 ARTLIST_MCP_TOKEN=
+# ARTLIST_AUTO_CONFIRM=false
 # Pick a specific model group. Two ways:
 #  - alias:   VIDEO_BACKEND=artlist_seedance-2.0  (also: -1.0-pro-fast, -1.5,
 #             -2.0-fast, -2.0-mini; plus kling-2.1, veo-3.1, sora-2, happyhorse-1.1)
--- a/.env.example
+++ b/.env.example
@@ -243,6 +243,18 @@ NEWAPI_VIDEO_GENERATE_AUDIO=auto
 # 支持 generate_audio 的视频模型。
 NEWAPI_VIDEO_AUDIO_MODELS=seedance-1.5-pro,seedance-2.0,seedance-2.0-fast,seedance-2.0-value,seedance-2.0-fast-value
 
+# --- Native Artlist generative video (image-to-video via Artlist MCP gateway) ---
+# Set VIDEO_BACKEND=artlist to route shot video through Artlist's generative
+# models. DramaClaw connects to Artlist's MCP server as a partner client,
+# uploads the first frame, generates, polls, and downloads.
+# NOTE: this needs a NON-INTERACTIVE partner token for mcp.artlist.io — the
+# per-user claude.ai connector cannot be used from a backend. Confirm the exact
+# partner auth scheme with Artlist; this client sends `Authorization: Bearer`.
+ARTLIST_MCP_TOKEN=
+# Optional overrides:
+# ARTLIST_MCP_URL=https://mcp.artlist.io/mcp
+# ARTLIST_VIDEO_MODEL_GROUP_ID=   # auto-routes to an i2v-capable model group
+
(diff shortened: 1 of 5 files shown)


lywaterman:
Thanks for exploring this integration. This is not ready to merge yet because the critical server-to-server contract has not been verified.

The public Artlist MCP flow uses interactive account authorization, while this implementation assumes an undocumented partner bearer token. The PR also notes that no end-to-end generation has succeeded with real credentials. We should not expose a production provider until authentication, tool names, request fields, confirmation flow, polling payloads, and downloads have all been exercised against the real service.

There is also a capability mismatch: aliases such as `artlist_seedance-2.0` imply Seedance 2.0 support, but this generator only forwards one first-frame image plus prompt/aspect/duration. It does not preserve the multi-image, audio, video-reference, resolution, or generated-audio capabilities expected by the existing Seedance workflow.

Please add contract tests with mocked MCP responses, a documented verified auth scheme, capability declarations/validation per model group, and update `license-inventory.csv` for the new source file before this can be reconsidered.

revolutionarybukhari:
Thanks — I've addressed the reviewable parts and been explicit about what can't be verified without live credentials. Pushed `2951e71`:

- **Mocked-MCP contract tests** (`tests/test_artlist_video.py`, 22): tool names + order (`upload_image`→`generate_video`→`get_generation_status`), request fields, the presigned upload→PUT→`confirm_upload` flow, the cost-confirmation flow (refused by default), polling payload parsing, and download success/failure.
- **Capability mismatch resolved honestly:** per-group `ModelCapabilities` declaration; requests using multi-image/keyframe, video refs, audio refs, or generated audio are now **rejected with a clear error** rather than silently dropped. Resolution is now actually forwarded. I did not claim Seedance omni parity — those capabilities stay declared-unsupported until the live contract is verified.
- `.env.example` rewritten as an explicit UNVERIFIED / not-production-ready warning; `ARTLIST_AUTO_CONFIRM` documented. Added the source file + test to `license-inventory.csv`.

**Still pending live verification (no partner credentials):** the `Bearer <partner-token>` auth against `mcp.artlist.io`, and a successful end-to-end generation. Tool names, exact field names (esp. `settings.resolution`/`confirmationId`/`confirm`), polling keys, and download URL location are pinned by mocks but not confirmed live — the tests guard the wiring only, and that's documented as such. I have not marked anything 'verified' that isn't. If you can share the partner auth scheme I'll exercise it against the real service.

lywaterman:
Thanks for the capability declarations and expanded tests. I re-reviewed the latest head, and this still cannot merge:

1. The file itself states that server-side auth, MCP tool names, request fields, confirmation flow, polling payloads, and download URLs have not been verified by a successful live run. Mock tests only prove that the guessed client and guessed responses agree with each other. We need an authoritative Artlist partner contract and a successful live smoke test before shipping this as a production backend.
2. There is still a silent input-drop bug. `generate()` accepts `image_path` as the first frame. If `references` contains exactly one additional image, `validate_capabilities()` computes `max(0, image_refs - 1) == 0` and allows it. But `_run()` uploads only `image_path` and never forwards `references`, so that extra reference is silently discarded. Any image in `references` is additional to `image_path` and must be rejected unless it is actually forwarded.
3. This also introduces a new direct provider credential/path outside NewAPI. Historical direct providers exist, but new production model integrations are expected to use NewAPI for consistent routing, billing, model aliases, and provider hiding.

Please resolve the input validation bug and rework the integration around a verified, production-supported contract—preferably through NewAPI.

revolutionarybukhari:
Closing after scoping the NewAPI-routing path you asked for.

Artlist's video generation is driven over **MCP tool-calling** (`upload_image` → presigned PUT → `confirm_upload` → `generate_video` → `get_generation_status`), which NewAPI's OpenAI-compatible HTTP video path (`/video/generations` + `/videos/{id}`) cannot proxy — NewAPI has no MCP-client channel type, so representing Artlist behind NewAPI would require a gateway-product change, not a config/alias in this repo.

Separately, every model this provider aliases (Seedance / Kling / Veo / Sora / HappyHorse) is **already reachable through NewAPI directly**, so the direct provider only adds an un-proxyable second transport with no new capability. Closing in favor of the existing NewAPI models. Thanks for the thorough reviews.

PR #232, opened 2026-08-01
fix(ingest): report live graph import progress

lywaterman:
## What changed

- Add granular graph-ingest progress milestones across source validation, Cognee ingestion, graph validation, vector indexing, preview generation, and final persistence.
- Preserve the last numeric progress value when Cognee emits log-only status messages instead of resetting the bar to zero.
- Show a live elapsed-time counter beside the import percentage while graph ingest is active.
- Add backend and frontend regression coverage for monotonic progress and the running timer.

## Why

Graph ingest can take several minutes. The previous flow often appeared stuck because ordinary log messages reset progress to 0 and the UI showed no indication that time was still advancing.

## Impact

Users now see meaningful intermediate progress plus an elapsed timer throughout long graph imports. Task completion behavior and imported project data are unchanged.

## Validation

- `uv run pytest tests/test_ingest_progress.py -q` — 1 passed
- `pnpm exec vitest run src/__tests__/components/ingest-elapsed-time.test.tsx src/__tests__/routes/ingest-settings-save.test.tsx` — 30 passed
- `pnpm exec tsc -b`
- `uv run ruff check src/novelvideo/cognee/store.py src/novelvideo/task_backend/runners/ingest.py tests/test_ingest_progress.py`
- `npx @google/design.md lint DESIGN.md`
- `git diff --check origin/main..HEAD`


Handanhhhy (dismissed):
P2 — Start the local elapsed-time fallback when the ingest task is submitted, not before settings save.

`frontend/src/routes/_app/projects.$project/ingest.tsx:1447` captures `localIngestStartedAtMs` before `await saveProjectSettings()`. If that request takes 60 seconds and the task-list/SSE row has not arrived yet, the import card immediately renders `Elapsed 01:00` even though the ingest task was just submitted. Move the fallback timestamp assignment to immediately before/after the ingest submission (using the actual submission boundary).

Red regression test: a delayed `updateProject` advances `Date.now()` from 1s to 61s, then starts ingest; expecting `Elapsed 00:00` fails because the PR renders `Elapsed 01:00`. Command: `pnpm exec vitest run src/__tests__/routes/ingest-settings-save.test.tsx -t "starts the local elapsed fallback when the ingest task is submitted, not before settings save"`.

PR #233, opened 2026-08-01
fix(sketch-edit): preserve task progress on logs

lywaterman:
## What changed

- preserve the latest task progress when sketch-edit executor emits ordinary log messages
- allow log-only task updates to pass `progress=None`
- add a regression assertion for the emitted progress sequence

## Why

Sketch-edit batch logs previously called the progress updater with `0.0`, which could make the task-center progress bar jump back to zero between real milestones.

## User impact

Sketch-edit execution now keeps its last reported percentage while still updating the current task message.

## Verification

- `uv run pytest tests/test_sketch_edit_execute_celery.py -q` (2 passed)
- `uv run ruff check src/novelvideo/task_backend/runners/sketch_edit_execute.py tests/test_sketch_edit_execute_celery.py`
- `git diff --check origin/main...HEAD`


Handanhhhy (dismissed):
Reviewed the task-progress update path and regression coverage. `progress=None` is handled by `TaskStateManager` as preserving the existing numeric progress while still updating task text/logs. The focused regression is red on the base behavior (`[0.01, 0.5, 0.0]`) and green on this PR (`[0.01, 0.5, None]`); focused tests and Ruff pass. No actionable findings.

Handanhhhy (approved):

PR #252, opened 2026-08-05
refactor(seedance2): return prompt as plain text

lywaterman:
## 变更
- 将 Seedance 2.0 Prompt Composer 从单字段结构化输出改为普通文本
- 不再为该纯文本任务关闭推理或强制 tool call
- 保留 prompt 规范化、空结果失败与规则草稿回退

## 验证
- uv run ruff check src/novelvideo/seedance2_i2v/prompt.py tests/test_newapi_text_gateway.py tests/test_seedance2_prompt.py
- uv run pytest -q tests/test_newapi_text_gateway.py tests/test_seedance2_prompt.py tests/test_api_seedance2_prompt_route.py tests/test_seedance2_panel_service_narration_audio.py
- 43 passed

## 影响
只调整 Seedance 2.0 提示词生成的 LLM 输出协议，不修改视频生成、计费或前端 API 合同。

### DIFF
--- a/src/novelvideo/seedance2_i2v/spoken_dialogue.py
+++ b/src/novelvideo/seedance2_i2v/spoken_dialogue.py
@@ -14,6 +14,9 @@ _SPEAKER_PREFIX_RE = re.compile(
     r"(?:（(?P<action>[^）]{0,120})）|\((?P<action_ascii>[^)]{0,120})\))?"
     r"\s*[：:]"
 )
+_LEADING_PERFORMANCE_NOTE_RE = re.compile(
+    r"^\s*(?:（[^）]{1,120}）|\([^)]{1,120}\))\s*"
+)
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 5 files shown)


src/novelvideo/seedance2_i2v/spoken_dialogue.py
```diff
@@ -65,10 +63,6 @@ def parse_seedance2_spoken_lines(beat: dict[str, Any]) -> list[Seedance2SpokenLi
     if lines:
         return lines
 
-    speaker = _text(beat.get("speaker"))
-    text = dialogue_text(beat)
-    if speaker and text:
-        return [Seedance2SpokenLine(speaker=speaker, text=text)]
     return []
```
Handanhhhy, line 71: [P1] Preserve validation for beat-level plain dialogue

When a dialogue beat uses audio_type=dialogue plus a beat-level speaker and plain narration_segment text (without an explicit 角色： prefix), this function now returns an empty list. _validate_dialogue_final_prompt therefore skips the dialogue-content check and only verifies the audio label. A final prompt that mentions 音频1 but omits the spoken line entirely passes validation and can proceed to video generation.

A minimal regression test expecting 缺少台词内容 fails with DID NOT RAISE ValueError on this commit. Please keep the raw text for the composer if needed, but retain an independent validation fallback for this common dialogue shape (for example via dialogue_text()).

Handanhhhy (approved):

PR #260, opened 2026-08-06
fix(audio): guard narrator voice prerequisites

lywaterman:
## What changed

- distinguish an unconfigured narrator voice from a configured but unreadable voice file
- block single-Beat, selected-Beat, and whole-episode audio submission when narrator prerequisites are missing
- hide misleading credit quotes and guide users to configure the required voice
- add backend and frontend regression coverage ensuring invalid requests are not enqueued

## Why

Narrated projects require a usable narrator voice source. Previously, missing configuration and missing storage files produced the same error, making deployment/storage faults look like user input errors. Some frontend audio entry points could also continue toward confirmation even when the quote API had already reported the missing prerequisite.

## User impact

- users receive an actionable configuration message before task creation or credit reservation
- configured-but-missing files now point operators toward project storage instead of asking users to configure the voice again
- first-person narration behavior remains unchanged and continues to use the protagonist identity voice

## Verification

- `uv run pytest tests/test_api_audio_prereq.py tests/test_api_audio_indextts2_cutover.py tests/test_indextts2_beat_audio_task.py tests/test_seedance2_voice_clone.py tests/test_project_config_narration.py tests/test_narrator_main_semantics.py tests/test_hermes_dramaclaw_plugin.py -q` — 89 passed
- focused frontend tests — 38 passed
- `npm run build`
- Ruff checks for changed Python files
- `git diff --check`
- `npx @google/design.md lint DESIGN.md` — 0 errors

### DIFF
--- a/frontend/src/__tests__/components/episode/beat-workbench/audio-pane.test.tsx
+++ b/frontend/src/__tests__/components/episode/beat-workbench/audio-pane.test.tsx
@@ -15,6 +15,7 @@ const i18n = i18next.createInstance();
 const mutateRegenerate = vi.hoisted(() => vi.fn());
 const toastError = vi.hoisted(() => vi.fn());
 const navigateMock = vi.hoisted(() => vi.fn());
+const audioPrereqErrors = vi.hoisted(() => ({ current: [] as string[] }));
 
 beforeAll(async () => {
   await i18n.use(initReactI18next).init({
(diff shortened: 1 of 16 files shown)


Handanhhhy (commented):
Two P2 issues to address:

1. The new frontend guards treat cached `prereq_errors` as authoritative, but narrator/character voice mutations do not invalidate the `audio-billing-quote` queries. Because the app has a global 30-second `staleTime`, a user who uploads or fixes the required voice and immediately returns can still see “configure voice” and remain blocked until the cache expires. Please invalidate the project's audio quote queries after voice mutations, or make these quotes refetch on mount.

2. The configured-but-unreadable diagnostic is still incomplete. `narrator_reference_audio_path()` only checks `exists()`, then `file_sha256()` opens the path without handling I/O errors. A configured path that resolves to a directory (or another read failure) raises and can turn the billing/generation request into a 500 instead of returning the intended storage diagnostic. Please validate that the resolved path is a readable file and handle `OSError` while hashing.

Handanhhhy (approved):

PR #273, opened 2026-08-07
fix video generation mode contract

ryanhuang-duat:
## Summary

- require the frontend to send an explicit video generation mode for every Freezone video route
- preserve the selected business mode separately from the provider execution protocol
- stop inferring first-frame versus first/last-frame behavior from the number of submitted images
- persist the selected mode in node history and add regression coverage for all keyframe combinations

## Root cause

The keyframe route inferred the execution mode from whether one or two images were present. As a result, selecting first/last-frame mode with only one connected image was silently reclassified as first-frame mode, which could conflict with the model capabilities configured in the media catalog.

## Impact

Admin-configured video capabilities, frontend mode selection, backend validation, billing metadata, task payloads, and history restoration now use one consistent business-mode contract. Provider-specific execution values remain internal to the task payload.

## Validation

- `uv run pytest -q tests/test_freezone_video_route_modes.py tests/test_freezone_image_backend.py tests/test_history_asset_model_mode.py tests/contract/test_m06_route_contracts.py` — 225 passed
- focused Ruff checks — passed
- frontend TypeScript check — passed
- `git diff --check origin/main...HEAD` — passed

### DIFF
--- a/frontend/src/api/ops.ts
+++ b/frontend/src/api/ops.ts
@@ -195,8 +195,8 @@ export interface FreezoneVideoGenPayload extends FreezoneNodeContext {
   generateAudio?: boolean;
   /** Backend model id, e.g. huimeng_seedance20_fast / seedance_pro. */
   model?: string;
-  /** 生成模式（还原用）：textToVideo / imageToVideo / firstLastFrame / imageReference / allReference。 */
-  genMode?: string;
+  /** 文生视频入口的固定业务模式。 */
+  genMode: "textToVideo";
   /**
    * Real-person material review. Set `true` when the input contains real
    * human faces so the backend routes the job through the human-review path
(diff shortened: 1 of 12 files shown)


Handanhhhy (changes requested):
Requesting changes for one correctness issue:

[P1] Normalize `firstFrame` before request-schema selection (`src/novelvideo/api/routes/freezone.py:7979`). The route now passes the business-mode value `firstFrame` to `media_request_schema_for_mode()`, but `MODE_ALIASES` has no `firstFrame -> first_frame` mapping. Consequently, every model parameter scoped to `first_frame` is silently removed; submitted values are dropped and required parameters are no longer validated.

Reproduction: a regression test calling `media_request_schema_for_mode(schema, "firstFrame")` with a `first_frame`-scoped `camera_fixed` parameter fails with `assert [] == ["camera_fixed"]`.

Please pass `execution_mode` here or add the missing alias with regression coverage.

The PR focused backend suite passes (225 tests), as do the frontend TypeScript check and `git diff --check`.

Handanhhhy (approved):

PR #278, opened 2026-08-07
feat(catalog): publish official media catalog via OSS

bopy-zou:
## Summary

- publish the official media catalog to dramaclaw-dl using immutable SHA256 object paths and a short-lived manifest
- verify catalog versions and hashes in the CE backend, revalidate with ETag, and expose publication status
- add an opt-in background poller, bilingual configuration docs, and a GitHub Actions publisher

## Verification

- uv run pytest tests/test_publish_official_media_catalog.py tests/test_model_gateway_settings.py -q (90 passed)
- uv run ruff check on changed Python modules and tests
- pnpm --dir frontend exec tsc -b
- uv run python scripts/check_env_config.py
- uv run python scripts/check_dependency_licenses.py
- publisher dry-run against the bundled catalog

## Configuration impact

The workflow targets oss-cn-chengdu.aliyuncs.com, bucket dramaclaw-dl, prefix official-media-catalog by default. It reads OFFICIAL_CATALOG_OSS_ACCESS_KEY_ID and OFFICIAL_CATALOG_OSS_ACCESS_KEY_SECRET from GitHub Actions secrets, with the existing OSS_RELAY_AK and OSS_RELAY_SK names as fallback. The bucket remains private; only anonymous GetObject on official-media-catalog/* is required.


Handanhhhy (changes requested):
**P1 — Publisher workflow fails before validation or upload**

The workflow installs only `oss2`, but `scripts/publish_official_media_catalog.py` imports `novelvideo.model_gateway_settings`, which transitively imports `novelvideo.env` and then `dotenv`. I reproduced the declared workflow environment in a clean virtualenv with `oss2==2.19.1`; `python scripts/publish_official_media_catalog.py --dry-run` exits with `ModuleNotFoundError: No module named dotenv`.

Please install the required project/runtime dependencies in `.github/workflows/publish-official-media-catalog.yml`, or move catalog validation into a dependency-free module. Add a regression test that invokes the publisher dry-run in an isolated interpreter containing only the dependencies declared by the workflow.

bopy-zou:
Addressed the P1 in 81e08eb6. The publisher now imports catalog validation from a third-party-free module, and the runtime settings code reuses the same validator. The regression test launches the publisher with Python isolated mode and site-packages disabled (-I -S), so the dry-run succeeds without dotenv or any other undeclared package; the upload path still imports oss2 lazily after validation. Post-rebase verification: 95 focused Python tests passed, Ruff passed, 3 catalog watcher frontend tests passed, and TypeScript compilation passed. Please re-review.

Handanhhhy (changes requested):
**P2 — Resolve catalog paths against the final manifest URL**

`check_official_media_catalog_update()` creates its client with redirects enabled, but `_check_with_client()` resolves `manifest["path"]` against the original `source_url`. If the configured manifest endpoint redirects to another host or directory (for example, `https://catalog.example.test/manifest.json` → `https://cdn.example.test/releases/manifest.json`), the code requests the catalog from `https://catalog.example.test/catalogs/<sha>.json` instead of `https://cdn.example.test/releases/catalogs/<sha>.json`, and the update fails with 404.

I reproduced this with an `httpx.MockTransport` regression test; the original focused suite still passes (95 tests), while the redirect test reliably fails at `catalog_response.raise_for_status()` with the incorrect original-host URL.

Please resolve the relative catalog path against the final manifest response URL (`response.url`) and add the redirect regression test.

Handanhhhy (approved):

PR #287, opened 2026-08-11
fix(config): distinguish text-model timeouts from real failures (Closes #211)

### LINKED ISSUE #211: bug: Freezone 长耗时文本生成在本地超时后丢失上游成功结果
### 前置检查 / Pre-flight checks

- [x] 我已搜索过现有 issue，确认这不是重复 / I searched existing issues and this isn't a duplicate
- [x] 我用的是最新版本，或已在下方写明所用 commit / I'm on the latest version, or stated the commit below

### 问题描述 / What happened

Freezone 故事脚本等结构化文本任务通过 OpenAI 兼容客户端调用模型时，默认受 `MODEL_TIMEOUT=120` 秒限制。当单次模型请求超过该时间，客户端抛出 `Request timed out`，任务随即被标记为失败。

但超时只代表 DramaClaw 不再等待响应，并不一定代表上游停止执行。上游可能随后成功完成该请求，而 DramaClaw 已结束本次任务，无法接收和写回成功结果，最终表现为：

- 用户看到任务失败；
- 上游实际可能已经完成生成；
- 已生成的结果无法恢复；
- 任务状态与上游最终状态不一致。

当前文本模型客户端在 `src/novelvideo/config.py` 中使用 `AsyncOpenAI(timeout=timeout_seconds)`，默认超时时间为 120 秒；长文本、复杂结构化输出及输出校验重试更容易触发该问题。

该问题与 #160 所描述的“任务已经入队，但提交响应在浏览器侧丢失”不是同一阶段：#160 发生在任务提交阶段，本问题发生在后台任务执行过程中的模型 HTTP 调用阶段。

积分预扣、确认和退款的统一结算由 #210 单独处理，不属于本 Issue 的修复范围。

### 复现步骤 / Steps to reproduce

1. 保持文本模型请求的默认超时配置（120 秒）。
2. 在 Freezone 提交一个输出较长、结构较复杂的故事脚本生成任务。
3. 让兼容 OpenAI 的上游在 120 秒后才完成响应。
4. 观察 DramaClaw 在约 120 秒时显示 `Request timed out`，并把任务标记为失败。
5. 检查上游请求状态，可观察到同一请求可能在客户端超时后成功完成。
6. 上游生成结果不会回写到原任务，任务也不会从失败状态恢复。

### 期望行为 / Expected behavior

- 后台生成任务不应仅因单次 HTTP 客户端达到固定的 120 秒超时，就进入不可恢复的失败状态。
- 文本模型调用超时应与任务整体截止时间、模型能力和网关超时策略保持一致，并支持按任务或模型配置。
- 如果本地停止等待但上游可能继续执行，应将状态标记为“结果待确认”并支持后续恢复，或者明确取消上游请求，避免产生无法交付的生成结果。
- 上游最终成功时，生成结果应能够关联并写回原任务。
- 超时错误应使用明确、可识别的错误码，并补充覆盖“客户端先超时、上游后成功”的回归测试。

### 部署方式 / Deployment

其他 / Other

### 版本或 commit / Version or commit

当前 main（问题代码路径可见于 `b71248e`）

### 运行环境 / Environment

服务端部署；通过 OpenAI 兼容模型网关调用长耗时文本模型。生产账号、项目、请求及渠道标识均已省略。

### 相关日志 / Relevant logs

```shell
Request timed out.
```

madebysaira:
## What

Fixes #211 (Freezone long-form text generation losing upstream success after a local client timeout).

When a text-model request runs longer than `MODEL_TIMEOUT` (default 120s), the OpenAI-compatible HTTP client raises a generic `Request timed out` and the background task is marked failed. That timeout only means DramaClaw stopped waiting — the upstream provider may still be completing the request, and the generated result is then lost with no way to tell a real failure from a timeout.

This PR makes that distinction possible and gives long-running tasks a way to opt out of the fixed global timeout:

- **`ModelTimeoutError`** — a dedicated, identifiable error (stable code + model name + timeout in the message) raised when a text-model call times out at the client boundary, instead of a generic message.
- **Per-task timeout override** — callers that know a task is long-running can pass an explicit `timeout_seconds_override` at the call site; existing callers keep today's behavior (env/gateway default).
- **Regression tests** — cover the "client times out first" path and confirm the raised error carries the expected code and metadata; unrelated errors (auth/config) still propagate unchanged.

Billing/recovery-state handling is intentionally out of scope (tracked separately in #210).

## Changes

- `src/novelvideo/config.py` — `ModelTimeoutError` + timeout translation at the client boundary + override plumbing
- `tests/test_model_timeout_error.py` — new regression tests

## How verified

- `ruff check` clean on touched files
- New tests pass (`pytest tests/test_model_timeout_error.py`)

Closes #211

### DIFF
--- a/src/novelvideo/config.py
+++ b/src/novelvideo/config.py
@@ -17,6 +17,143 @@ from novelvideo.official_defaults import (
 load_dotenv()
 
 # =============================================================================
+# 文本模型超时错误
+# =============================================================================
+
+MODEL_TIMEOUT_ERROR_CODE = "MODEL_TIMEOUT"
+
+
+class ModelTimeoutError(RuntimeError):
+    """Client-side timeout while waiting for a text model.
+
+    The local client gave up after ``timeout_seconds`` seconds while
+    contacting ``model_name``. The upstream may still complete
+    successfully — this error is distinguishable from real model
+    failures via ``error_code == "MODEL_TIMEOUT"``.
+    """
+
+    error_code = MODEL_TIMEOUT_ERROR_CODE
+    code = MODEL_TIMEOUT_ERROR_CODE
+
+    def __init__(
+        self,
+        model_name: str,
+        timeout_seconds: float,
+        message: str | None = None,
+    ) -> None:
+        self.model_name = str(model_name or "").strip() or "unknown"
+        self.model = self.model_name
+        self.timeout_seconds = float(timeout_seconds)
+        self.error_code = self.error_code
+        self.code = self.code
+        if message is None:
+            message = (
+                f"text model '{self.model_name}' timed out after "
+                f"{self.timeout_seconds:g}s (client timeout; "
+                f"upstream may still succeed) [{self.error_code}]"
+            )
+        super().__init__(message)
+
+
+def _is_text_model_timeout_error(exc: BaseException) -> bool:
+    """Return True when *exc* is a client-side timeout.
+
+    Recognised signals:
+    - ``asyncio.TimeoutError``
+    - ``httpx.TimeoutException`` (and subclasses)
+    - ``openai.APITimeoutError``
+    - ``openai.APIConnectionError`` whose message/cause indicates timeout
+    """
+
+    try:
+        import asyncio as _asyncio
+
+        if isinstance(exc, _asyncio.TimeoutError):
+            return True
+    except Exception:
+        pass
+
+    try:
+        import httpx as _httpx
+
+        if isinstance(exc, _httpx.TimeoutException):
+            return True
+    except Exception:
+        pass
+
+    try:
+        from openai import APIConnectionError as _APIConnectionError
+        from openai import APITimeoutError as _APITimeoutError
+
+        if isinstance(exc, _APITimeoutError):
+            return True
+        if isinstance(exc, _APIConnectionError):
+            msg = str(exc).lower()
+            if "timeout" in msg or "timed out" in msg:
+                return True
+            cause = getattr(exc, "__cause__", None)
+            if cause is not None:
+                try:
+                    import httpx as _httpx2
+
+                    if isinstance(cause, _httpx2.TimeoutException):
+                        return True
+                except Exception:
+                    pass
+                try:
+                    import asyncio as _asyncio2
+
+                    if isinstance(cause, _asyncio2.TimeoutError):
+                        return True
+                except Exception:
+                    pass
+                if "timeout" in str(cause).lower():
+                    return True
+            ctx = getattr(exc, "__context__", None)
+            if ctx is not None:
+                try:
+                    import httpx as _httpx3
+
+                    if isinstance(ctx, _httpx3.TimeoutException):
+                        return True
+                except Exception:
+                    pass
+                if "timeout" in str(ctx).lower():
+                    return True
+            return False
+    except Exception:
+        pass
+
+    return False
+
+
+def get_newapi_text_timeout_seconds(
+    model_env: str,
+    *,
+    timeout_seconds_override: float | None = None,
+    default_seconds: float = 300.0,
+) -> float:
+    """Resolve the effective text-model timeout for *model_env*.
+
+    Priority:
+    1. explicit ``timeout_seconds_override`` (per-task/per-call)
+    2. ``{model_env}_TIMEOUT_SECONDS`` env var
+    3. ``NEWAPI_TEXT_TIMEOUT_SECONDS`` env var
+    4. *default_seconds* (300s by default)
+    """
+
+    if timeout_seconds_override is not None:
+        return float(timeout_seconds_override)
+    return _env_float(
+        f"{model_env}_TIMEOUT_SECONDS",
+        _env_float("NEWAPI_TEXT_TIMEOUT_SECONDS", float(default_seconds)),
+    )
+
+
+# Alias kept for internal callers that expect ``resolve_*`` naming.
+resolve_newapi_text_timeout_seconds = get_newapi_text_timeout_seconds
+
+# =============================================================================
 # 模型提供商配置
 # =============================================================================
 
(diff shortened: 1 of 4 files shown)


lywaterman (changes requested):
_(Reposting in English — the original was in Chinese, which doesn't match this PR. Content is unchanged.)_

**Request changes** — one substantive issue: `ModelTimeoutError` is never wired into the task-failure classifier, so the goal of "make timeouts distinguishable" doesn't hold end to end. The implementation details themselves are careful; separating the two below.

## ✅ Implementation details I verified individually — all correct

- **Timeout detection is narrow.** `asyncio.TimeoutError`, `httpx.TimeoutException`, `openai.APITimeoutError`, plus `APIConnectionError` **only when** `__cause__` / `__context__` is a timeout. Every import is `try`-wrapped, so a missing library doesn't take the path down, and non-timeout connection errors aren't misclassified as timeouts.
- **The `except BaseException` is safe.** Already-a-`ModelTimeoutError` re-raises bare (no double wrapping), non-timeouts re-raise bare. So `asyncio.CancelledError` reaches the final `raise` and **structured-cancellation semantics are preserved** — catching `BaseException` without that final bare `raise` is the usual trap here, and it's avoided.
- **Both the CE and EE paths are covered.** The translation lives in `_AutoClosingOpenAIChatModel` and wraps both `request` and `request_stream`; `_newapi_text_openai_model` is exactly the `delegate_factory` passed to `create_request_scoped_gateway_model(...)` on the EE branch, so org-path calls get `ModelTimeoutError` too.
- **No existing timeout handling is disturbed.** Every other timeout-typed branch in the repo (`chat/hermes_sdk.py:335/486/655`, `chat/backend_sdk.py:805`, `sqlite_store.py:525`) wraps its own `wait_for` and is unrelated to text-model calls. `task_backend/run_core.py:389` tests `isinstance(exc, TaskTimedOut)` — a concrete type — so `ModelTimeoutError` won't match and **won't be misreported as "task exceeded N minutes."** I checked this specifically, because `ModelTimeoutError` happens to also carry a `timeout_seconds` attribute; a duck-typed check there would have misfired. It isn't duck-typed, so we're fine.

## ❌ The one real issue: `error_code` is dropped on the way out

`MODEL_TIMEOUT` appears **only in `config.py` — it has zero consumers** (0 occurrences elsewhere in `src` and in `frontend/src`):

```
config.py:23  MODEL_TIMEOUT_ERROR_CODE = "MODEL_TIMEOUT"
config.py:35  error_code = MODEL_TIMEOUT_ERROR_CODE
config.py:36  code = MODEL_TIMEOUT_ERROR_CODE
```

The reason is that `_project_task_failure_for_exception` (`task_backend/run_core.py:382`) is an **explicit `isinstance` allowlist**: `NovelImportRequiredError`, `TaskTimedOut`, `SoftTimeLimitExceeded`, `Sharp3DUnavailable`, `BlockWorldUnavailable`… each branch explicitly puts `exc.error_code` into the metadata. `ModelTimeoutError` isn't in the list, so it falls through to the last line:

```python
return safe_exception_message(exc), {}, False
```

**The metadata is an empty dict**, so `error_code` is lost. The task record never carries `MODEL_TIMEOUT`, and the frontend still can't tell "upstream may have completed, we just stopped waiting" from a genuine failure — which is the thing #211 asks for.

What actually improves today is the **message text** (it now names the model and the timeout). The distinguishability is one step short.

### Fix

One branch in `_project_task_failure_for_exception`:

```python
if isinstance(exc, ModelTimeoutError):
    return str(exc), {
        "error_code": exc.error_code,
        "model": exc.model_name,
        "timeout_seconds": exc.timeout_seconds,
    }, True
```

The third return value is the terminal-failure flag — please pick it deliberately. A timeout means **upstream may in fact have succeeded**, so marking it a definite failure may be the wrong call. Worth spelling out the trade-off in the PR, since it's the crux of #211.

Also, while you're there: **please add a test** asserting that a timeout, run through `_project_task_failure_for_exception`, yields metadata with `error_code == "MODEL_TIMEOUT"`. The existing 399 lines cover "the translation layer raises an error with the right attributes," but nothing covers "the metadata survives all the way to the task record" — and that gap is precisely why the issue above went unnoticed.

## ⚠️ The PR description presents an existing capability as new

This reads like a new interface:

> **Per-task timeout override** — callers that know a task is long-running can pass an explicit `timeout_seconds_override`

But it **already exists on `main`** (4 occurrences in `config.py`) and is already in use:

```
freezone/text_node.py:570       timeout_seconds_override=300.0
freezone/vision_gateway.py:67   timeout_seconds_override=timeout_seconds
```

What this PR does is refactor the resolution into a priority-ordered helper (explicit override → env/gateway default), which is fine — but please reword the description. Otherwise a reviewer goes looking for a new interface that isn't there, or assumes existing call sites need updating.

## Two small ones

1. `self.error_code = self.error_code` and `self.code = self.code` in `ModelTimeoutError.__init__` are no-ops (assigning an instance attribute from the identically-valued class attribute) — just delete them.
2. `request_stream`'s `try` encloses the `yield`, so a timeout raised by the consumer inside the `async with` body would also be attributed to "text model timeout." In practice the consumer is reading the upstream stream, so the attribution is usually right and the risk is low — noting it for the record.

## Other

The branch is currently **BEHIND**; please rebase onto latest `main` before merging.

---

Good to merge once `ModelTimeoutError` is in the allowlist and that end-to-end assertion exists. The direction and the implementation are both right — it's just the last bit of wiring.

madebysaira:
## Address review feedback

- Wire `ModelTimeoutError` into the task-failure projection so `MODEL_TIMEOUT`, model, and timeout metadata reach the task record.
- Add an end-to-end regression test for that metadata path.
- Remove redundant instance assignments in `ModelTimeoutError`.

Targeted verification: `uv run pytest -q tests/test_task_credit_errors.py tests/test_model_timeout_error.py` — 26 passed.

The branch was rebased from the reviewed PR head and updated in place.

madebysaira:
Friendly re-review request — the review feedback has been addressed on this branch:

- `ModelTimeoutError` is now wired into the task-failure classifier (`_project_task_failure_for_exception`), so `MODEL_TIMEOUT`, model name, and timeout metadata land in the task record instead of being dropped as an empty dict.
- End-to-end regression test added covering the metadata path.
- Redundant instance assignments in `ModelTimeoutError` removed.

Verified with `uv run pytest -q tests/test_task_credit_errors.py tests/test_model_timeout_error.py` — 26 passed. Ready for another look whenever you have a moment.

madebysaira:
Rebased onto current `main` (`f7da95c7`) and resolved conflicts in `config.py` / `run_core.py` / `test_task_credit_errors.py` (upstream refactored timeout resolution and prerequisite imports since this branch was cut; the fix now layers on top of the new structure). Verified: `uv run pytest -q tests/test_task_credit_errors.py tests/test_model_timeout_error.py` — 33 passed (26 original + 7 new upstream tests). PR is mergeable again.

PR #292, opened 2026-08-13
fix NewAPI media request protocol

ryanhuang-duat:
## 背景

图片和视频请求在多个入口中存在字段位置、枚举大小写和自动参数语义不一致的问题；视频编辑使用自动时长后，前端报价与后端预扣也需要保持同一计费口径。

## 主要改动

- 统一图片、视频发往 NewAPI 的媒体协议：
  - 固定比例保留 `metadata.ratio` / `metadata.resolution`，并发送同源的 `width` / `height`
  - 自动比例统一为 `auto`，不再发送固定尺寸
  - 分辨率枚举统一为小写
  - 统一首帧、尾帧和参考媒体字段
- 视频编辑统一发送自动比例与自动时长：
  - `ratio: auto`
  - `duration: auto`
  - 前端报价与后端预扣均按输入视频实际整秒时长计算
- 首尾帧模式强制使用自动比例。
- 视频编辑支持独立参考音频，并按媒体目录能力校验素材数量和时长。
- 模型的原生音频能力由后台目录声明，用户仍可在支持时选择是否生成音频。
- 全能参考仅输入一个视频时，在最终请求提示词中明确这是视频参考生成新视频，不是视频编辑。
- 分辨率在 DramaClaw 界面中统一以小写显示。
- 保留模型专用可选参数机制，未选择的参数不会发送。

## 影响

- 不包含数据库迁移。
- 不修改促销、预扣、确认或退款流程。
- 计费规则的唯一语义变化是视频编辑 `duration=auto` 时，报价与预扣使用输入视频时长；其他计费单位与公式保持不变。

## 验证

- 后端定向测试：403 passed
- 前端定向测试：5 files / 110 tests passed
- `bun run build:ce`：passed
- `git diff --check origin/main...HEAD`：passed

### DIFF
--- a/frontend/src/__tests__/api/freezone-video-edit.test.ts
+++ b/frontend/src/__tests__/api/freezone-video-edit.test.ts
@@ -1,0 +1,44 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import { apiCall } from "@/api/client";
+import { submitFreezoneVideoEdit } from "@/api/ops";
+
+vi.mock("@/api/client", () => ({
+  apiCall: vi.fn(),
+  apiClient: {},
+}));
+
+describe("submitFreezoneVideoEdit", () => {
+  beforeEach(() => {
+    vi.mocked(apiCall).mockReset();
+    vi.mocked(apiCall).mockResolvedValue({ job_id: "job-1" });
+  });
+
+  it("forwards references while leaving geometry and duration to the source video", async () => {
+    await submitFreezoneVideoEdit("project-1", {
+      videoUrl: "/static/source.mp4",
+      imageUrls: ["/static/style.png"],
+      audioUrls: ["/static/music.mp3"],
+      model: "catalog-video",
+      genMode: "videoEdit",
+    });
+
+    expect(apiCall).toHaveBeenCalledWith(
+      "projects/project-1/freezone/video/video-edit",
+      expect.objectContaining({
+        method: "POST",
+        json: expect.objectContaining({
+          video_url: "/static/source.mp4",
+          image_urls: ["/static/style.png"],
+          audio_urls: ["/static/music.mp3"],
+          gen_mode: "videoEdit",
+        }),
+      }),
+    );
+    const request = vi.mocked(apiCall).mock.calls[0]?.[1] as {
+      json?: Record<string, unknown>;
+    };
+    expect(request.json).not.toHaveProperty("aspect_ratio");
+    expect(request.json).not.toHaveProperty("duration_seconds");
+  });
+});
--- a/frontend/src/__tests__/features/canvas/catalog-image-model-request.test.ts
+++ b/frontend/src/__tests__/features/canvas/catalog-image-model-request.test.ts
@@ -81,6 +81,14 @@ beforeEach(() => {
 });
 
 describe("目录模型的 provider/model 往返", () => {
+  it("小写分辨率目录仍默认选择 2k 档位", () => {
+    const definition = toImageModelDefinition(
+      entry({ resolutionOptions: ["1k", "2k", "4k"] }),
+    );
+
+    expect(definition.defaultResolution).toBe("2k");
+  });
+
   it("OpenRouter 命名空间模型：provider 归 openrouter，model 保留完整多段名", async () => {
     const body = await submittedBody(entry({}));
     expect(body.provider).toBe("openrouter");
--- a/frontend/src/__tests__/features/canvas/media-model-options.test.ts
+++ b/frontend/src/__tests__/features/canvas/media-model-options.test.ts
@@ -1,0 +1,30 @@
+// SPDX-License-Identifier: Elastic-2.0
+// Copyright (c) 2026 ClaymoreLab
+
+import { describe, expect, it } from 'vitest';
+
+import {
+  formatResolutionLabel,
+  resolveModelAspectOptions,
+  resolveModelSizeOptions,
+} from '@/features/canvas/domain/mediaModelOptions';
+
+describe('媒体模型公共选项归一化', () => {
+  it('把历史 adaptive 比例收敛为唯一的 auto 选项', () => {
+    expect(
+      resolveModelAspectOptions({ ratioOptions: ['adaptive', 'Auto', '16:9'] }),
+    ).toEqual(['auto', '16:9']);
+  });
+
+  it('保留后台配置的分辨率档位顺序', () => {
+    expect(resolveModelSizeOptions({ resolutionOptions: ['720p', '1080p'] })).toEqual([
+      '720p',
+      '1080p',
+    ]);
+  });
+
+  it('分辨率只在显示层统一成小写', () => {
+    expect(formatResolutionLabel('4K')).toBe('4k');
+    expect(formatResolutionLabel('1080P')).toBe('1080p');
+  });
+});
--- a/frontend/src/__tests__/features/canvas/video-model-capabilities.test.ts
+++ b/frontend/src/__tests__/features/canvas/video-model-capabilities.test.ts
@@ -19,7 +19,10 @@ import {
   referenceDurationLimitsMs,
   resolveVideoKeyframeUrls,
   videoEmptyStateCtaModes,
+  videoModeForcesAutomaticAspectRatio,
   videoModeRequiresPrompt,
+  videoModelDefaultGenerateAudio,
+  videoModelSupportsGenerateAudio,
   videoModelReferenceDisabledReason,
   videoMultiImageAutoSwitchMode,
   videoReferenceAutoSwitchAction,
@@ -27,6 +30,20 @@ import {
   videoSubmitMediaRejectionReason,
   videoUpstreamImageDefaultMode,
 } from "@/features/canvas/nodes/shared/videoModelCapabilities";
+
+describe("视频模式有效比例", () => {
+  it.each([
+    ["firstFrame", true],
+    ["firstLastFrame", true],
+    ["videoEdit", true],
+    ["textToVideo", false],
+    ["imageToVideo", false],
+    ["imageReference", false],
+    ["allReference", false],
+  ] as const)("%s 是否强制跟随输入素材", (mode, expected) => {
+    expect(videoModeForcesAutomaticAspectRatio(mode)).toBe(expected);
+  });
+});
 
 // 对齐后端 freezone/video_node.py 的真实模型 id（apiModel == id == 后端 model）。
 const SEEDANCE2_FAST = "newapi_seedance-2.0-fast";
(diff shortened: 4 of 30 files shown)


lywaterman (commented):
Reviewed together with the three sibling PRs (claymorelab/SuperTale#54, claymorelab/supertale-admin-fe#22, claymorelab/claymore-llm-gateway#38) since they form one wire-protocol change set.

**The cross-repo contract lines up.** I traced each dimension end to end:

| Dimension | Sent here | Consumed in gateway#38 |
|---|---|---|
| Auto duration | `duration: "auto"` | `TaskSubmitReq.UnmarshalJSON` → `DurationAuto` → `-1` for huimeng/tokenhub, omitted for happyhorse |
| Auto ratio | `metadata.ratio="auto"`, no `width`/`height` | `auto→adaptive` for seedance, cleared for happyhorse |
| Fixed ratio | `metadata.ratio` + same-source `width`/`height` | `ValidateAspectRatioDimensions` checks consistency; auto and width/height are mutually exclusive |
| Lowercase resolution | `normalize_media_resolution_value` | SuperTale normalizes on both read and write, so pre-existing uppercase rows still match |

Below are the findings scoped to this repo.

---

### 1. `SeedanceVideoGenerator` was not updated — first-frame ratio silently falls back to 9:16

`src/novelvideo/generators/video_generator.py:477`

```python
ratio = ratio_text if ":" in ratio_text or ratio_text == "adaptive" else "9:16"
```

The identical guard in `NewApiVideoGenerator` was updated in this PR (`video_generator.py:2581` → `ratio_text in {"auto", "adaptive"}`), but this direct-Seedance path still only accepts `adaptive`.

Since `freezone_video_keyframes` now hard-codes `aspect_ratio="auto"` (`src/novelvideo/api/routes/freezone.py:8366`), any project still on the direct `seedance_fast` backend takes the `else` branch and gets 9:16 — dropping exactly the follow-the-first-frame semantics this PR is meant to establish.

`HuimengVideoGenerator` (`video_generator.py:1545`, `else "adaptive"`) and `_video_dimensions` (`video_generator.py:2167`, where `auto` is caught by the `":" not in ratio_text` guard) are both fine — this is the only miss I found.

### 2. `video_edit` source video is constrained by the reference-clip duration bounds

`src/novelvideo/api/routes/freezone.py:405-438`

`_catalog_reference_duration_bounds(capabilities, "video")` is applied to every `type == "video"` entry in `reference_items`, which now includes the `video_edit` source video. Those bounds exist to cap the short clips fed to omni-reference, and the same catalog entry serves both modes.

A model configured with e.g. `referenceVideoMaxSeconds: 10` will 400 when a user submits a 30s video for editing. Suggest either scoping the bound to non-source reference videos, or adding a separate catalog field for the editable source length.

### 3. `video_edit` billing floor systematically under-collects vs. the gateway

`src/novelvideo/api/routes/freezone.py:447` and the frontend counterparts (`VideoNode.tsx`, `VideoOperationsPanel.tsx`) all use `floor`, and the comment states this is deliberate. Front and back end agree, so quote and pre-deduction are consistent.

The mismatch is one layer up: huimeng's and tokenhub's new `TaskBillingCompletionFacts` (gateway#38) settle on the **actual floating-point** duration returned upstream. For a 5.7s source, the gateway charges DramaClaw for 5.7s while DramaClaw charged the user for 5s. That is ~0.5s of margin lost per edit on average.

Not a bug if it is a deliberate product call — just flagging that the two ledgers now round in opposite directions.

### 4. Minor: `enforce_newapi_media_geometry_contract` can drop a fixed ratio entirely

`src/novelvideo/media_model_request_schema.py:164`

When the ratio is fixed but `width`/`height` are absent, neither branch runs. By that point top-level `ratio`/`aspect_ratio` have already been popped without being written back into `metadata`, so the ratio can be lost outright; an existing `metadata["ratio"]` also skips the lowercase normalization applied on the other paths.

I could not reach this through the current call chain (both the image and video paths always populate `metadata["ratio"]` first), so this is latent rather than live — but it is a sharp edge for the next caller.

ryanhuang-duat:
感谢联动四仓核对。这里按评论中的编号同步处理结论：

### 已处理

1. 旧直连 `SeedanceVideoGenerator` 现在同时接受 `auto` 和 `adaptive`，不会再把 canonical `auto` 静默回退为 `9:16`。当前 Freezone 目录路径通常已经映射到 NewAPI，这一处主要用于收口仍可能使用火山直连生成器的旧路径。

4. `enforce_newapi_media_geometry_contract()` 现在只要收到固定比例，就会始终保留为 `metadata.ratio`，不再要求同时存在 `width` / `height`。`auto` 路径仍会删除冲突尺寸，现有互斥协议不变。

对应回归测试已补充；定向执行 `tests/test_media_model_request_schema.py` 与 `tests/test_freezone_video_backend.py` 共 111 个测试通过，Ruff 通过。

### 按当前产品规则不调整

2. `referenceVideoMin/MaxSeconds` 和 `referenceVideoTotalMin/MaxSeconds` 在当前产品定义中是通用的“输入素材时长限制”，管理后台文案也明确为“分别控制单条素材和本次请求全部素材的时长”，并非仅用于 omni-reference 的短片限制。因此视频编辑源视频同样受该限制，是当前预期行为。未来若需要同一模型对全能参考和视频编辑设置不同上限，会另行增加按模式覆盖配置，而不是在本次请求链路中绕过限制。

3. 视频编辑按 `floor(input_video_duration_seconds)` 计入整数秒，是已经确认并完成前后端对齐的用户积分规则。Gateway 保留上游实际浮点时长用于供应商成本核算；两者属于用户积分账与上游成本账的不同口径。当前接受小数部分作为定价毛利成本，不在本 PR 中改成 `ceil` 或浮点积分，避免改变已确认的报价规则。

修复提交：`3296f95`。

lywaterman (commented):
Thanks for the detailed writeup. Re-checked against `3296f95`; CI is green (10/10). The three items you declined are all coherent product/architecture calls — the reference-duration bounds being a general input-material limit, `floor` being the confirmed user-credit rule with the fractional part booked as gross margin against the vendor cost ledger, and the fixed-ratio metadata now surviving without `width`/`height`. Confirming those was exactly the point of raising them; no further objection from me.

**Item 1 is only fixed at one of three sites, though.**

`_normalize_aspect_ratio` was added as a static method on `SeedanceVideoGenerator` and wired in at `video_generator.py:484`. Two identical colon-only guards remain in the same file (line numbers are on `3296f95`):

```python
# src/novelvideo/generators/video_generator.py:856   Seedance2VideoGenerator
ratio = aspect_ratio if ":" in aspect_ratio else "9:16"

# src/novelvideo/generators/video_generator.py:1132  GrokVideoGenerator
ratio = aspect_ratio if ":" in aspect_ratio else "9:16"
```

These are the same tier of legacy direct backend as the one you fixed — adjacent branches of the same factory:

| `video_generator.py` | `VideoBackend` | Generator | `auto` handling |
|---|---|---|---|
| `:3729` | `SEEDANCE_FAST` | `SeedanceVideoGenerator` | fixed in `3296f95` |
| `:3739` | `SEEDANCE_2` | `Seedance2VideoGenerator` | still falls back to `9:16` |
| `:3743` | `GROK_720` | `GrokVideoGenerator` | still falls back to `9:16` |

Both remaining sites accept first-frame input, so the keyframes route reaches them:

- `:856` is immediately preceded by `# 兼容旧接口：如果传了 image_path 但没有 references，自动作为首帧`, which promotes `image_path` to a 首帧 `ShotReference`.
- `:1132` rejects only the 尾帧 case (`Grok 720 does not support keyframe/首尾帧模式` at `:1126`) and accepts `image_path` as a first frame below.

Since `freezone_video_keyframes` now hard-codes `aspect_ratio="auto"`, both paths take the `else` branch and silently produce 9:16 — the exact behavior this PR set out to eliminate.

Worth noting `:856` is the worse of the two: that guard does not accept `adaptive` either, so the direct Seedance 2.0 generator has been converting `adaptive` → `9:16` all along. That is verbatim the defect described in the comment you removed from `:477` ("The old colon-only guard accidentally converted it back to 9:16") — it just also exists one class down.

Your stated rationale for fixing `:477` — 收口仍可能使用火山直连生成器的旧路径 — applies unchanged to both.

**Suggested shape:** hoist `_normalize_aspect_ratio` from `SeedanceVideoGenerator` to `VideoGeneratorBase` (or to module scope alongside the other normalizers) and call it from all three sites, so the next direct generator added cannot reintroduce the same guard. The `NewApiVideoGenerator` site at `:2589` already has the correct `{"auto", "adaptive"}` set and could share it too.

Everything else on my side is resolved.

ryanhuang-duat:
感谢继续核对。这里再沿实际运行调用链检查了一遍，这两处剩余的 colon-only guard 不属于当前受支持产品链路上的可触发问题，因此本 PR 不再继续给旧直连生成器补协议兼容。

当前虾画链路是：

```text
catalog 模型 / 历史选择值
→ _resolve_catalog_video_backend()
→ resolve_freezone_video_backend()
→ newapi_* backend
→ NewApiVideoGenerator
```

现有旧名称也已经在 `LEGACY_FREEZONE_VIDEO_BACKEND_ALIASES` 中重定向，例如：

```text
seedance_2       → newapi_seedance-2.0-fast
seedance_fast    → newapi_seedance-1.0-pro-fast
seedance_pro     → newapi_seedance-1.5-pro
```

因此 `freezone_video_keyframes` 传入 `aspect_ratio="auto"` 后，当前正式配置会进入已经正确处理 `auto/adaptive` 的 `NewApiVideoGenerator`，不会进入 `Seedance2VideoGenerator` 或 `GrokVideoGenerator`。当前主线默认视频 backend 同样已经是 `newapi_*`。

剩余两个类属于仍保留在工厂中的旧直连实现；只有绕开当前模型目录、显式提交旧 backend，或恢复未经转换的历史任务时才可能触发。继续为它们复制新协议规则会延长双链路维护，并不能解决根本问题。

后续会把这件事作为独立的遗留链路清理处理：保留旧 backend 名称到 `newapi_*` 的兼容映射，删除旧供应商直连实现及其工厂分支。这样既不破坏历史配置，也避免未来继续在两套请求协议上重复修复。

所以本次结论是：

- 当前 NewAPI 产品链路中的 `auto/adaptive` 问题已经修复；
- 两处旧直连 guard 不作为 #292 的阻塞项；
- 不在本 PR 扩大到旧直连代码重构或删除。

lywaterman:
You are right, and my reachability claim in the previous comment was wrong. Correcting it for the record.

I traced the chain you described and it holds. `aspect_ratio="auto"` is hard-coded in exactly two places repo-wide — `freezone.py:8367` (keyframes) and `freezone.py:8652` (video-edit) — and both resolve their backend through `resolve_freezone_video_backend`, which is a closed whitelist ending in `raise ValueError(f"unknown video model: {text}")`:

- `seedance_2` / `seedance_fast` / `seedance_pro` / `seedance_pro_silent` are all aliased to `newapi_*` in `LEGACY_FREEZONE_VIDEO_BACKEND_ALIASES` (`freezone/video_node.py:88-99`).
- `grok_720` appears in neither alias map and is not a `newapi_` name, so it raises. `newapi_grok-video-channel` is additionally listed in `FREEZONE_DISABLED_VIDEO_BACKENDS` (`video_node.py:120`).
- The default is `FREEZONE_DEFAULT_VIDEO_BACKEND = "newapi_seedance-2.0-fast"` (`video_node.py:110`).

So `Seedance2VideoGenerator` and `GrokVideoGenerator` cannot receive `auto` from any current path. My error was stopping at the factory branches in `video_generator.py` without following backend resolution one layer up.

One consequence worth recording, since it cuts the same way: under this reasoning the `:477` fix already merged in `3296f95` is also unreachable from the freezone route, so the rationale given for it (收口仍可能使用火山直连生成器的旧路径) was optimistic in the same direction. Not suggesting a revert — it is merged, covered by tests, and harmless as a defensive normalization. Just so the reachability story is consistent for whoever picks up the legacy cleanup.

Agreed that the follow-up shape you describe — keep the legacy backend-name aliases, delete the direct-provider implementations and their factory branches — is the right resolution rather than replicating the protocol rules into a second chain.

Nothing outstanding from me on this PR.

lywaterman (approved):

PR #293, opened 2026-08-13
fix(canvas): 低缩放档拖入文件不再留下一个空的上传节点

n7s4:
## 现象

低缩放档（zoom < 0.35）从外部拖入图片，两个症状：

1. 视口不会自动拉近到 0.6；
2. 放大后发现加进来的是一个空的「上传资源」节点，跟拖进去的图毫不相干。

## 根因

第二个是 LOD（#241）落地后的回归。投递链是「先 `addNode`、下一帧再把 `File` 发到 `canvasEventBus`」，而低缩放档下新节点先以 **LOD shell** 挂载，完整组件要等 `requestShellUpgrade` 的升级队列放行才挂上订阅 —— 必然晚于投递那一帧，而总线**无重放语义**，`File` 当场就丢了。

取证（真机 CDP）：拖入 1.5s 后节点仍是 `isShell:true / hasImg:false`，且 `[upload-perf][e2e] preview-state-committed` 一条都没有 —— `processFile` 根本没跑到。

第一个是纯漏调用：`handleCanvasDrop` 的文件分支和粘贴那条路都没走 `focusNewNodeIfLowZoom`，「凭空新建节点的入口在低缩放档拉近到 ≥0.6」这条约定对它们不生效。

## 改法

把 File 的载体与订阅时机解耦，不再依赖「一帧之内订阅得上」：

- 新增 `application/pendingExternalFiles.ts`：模块级暂存，`stashExternalFile` + **消费性**的 `takeExternalFile`（64 条上限兜底，防节点被删后条目常驻）；
- 新增 `hooks/useExternalFileHandoff.ts`：挂载时补投一次 + 收到事件再取一次。take 是消费性的，所以同一个 File 只会被处理一次，不需要额外去重标记；
- 三个投递方（画布拖入、粘贴、`spawnExternalAssetNodes`）改成 stash + 照旧发事件，事件降级成「敲一下已挂载的节点」；三个消费方（Upload / Video / Audio）换成这个 hook；
- 顺带覆盖 `UploadNode` 变形成 video/audio 的两处同构问题（拖入视频会走那条路）；
- 拖入与粘贴补上 `focusNewNodeIfLowZoom`。

## 验证

- 新增 `external-file-handoff-low-zoom.test.tsx`（4 例），接收侧用的是生产代码里的 hook 而非桩；把「挂载补投」那行摘掉后 2 例转红，确认钉的是这个 bug；
- `tsc -b` + `pnpm build` 干净，`src/__tests__/features/canvas` + `src/__tests__/stores` 全绿（105 文件 / 732 例）；
- 真机（EE 本地栈，20% 缩放拖入 PNG）：`zoom 0.2 → 0.6`，新节点标题为文件名、`img` 指向已上传的 `_uploads/...png`，无控制台报错；
- 新增的 3 个被跟踪文件已登记进 `license-inventory.csv`。

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lywaterman (commented):
The core design is right. `stashExternalFile` lands synchronously after `addNode` while `publish` is deferred to a rAF, so the consumer's mount-time `drain()` is guaranteed to see the file regardless of when — or whether — the LOD shell upgrades. Making `takeExternalFile` consuming is what removes the need for a separate dedup flag, and the new test exercises the production hook rather than a stub. Verified all five publish sites are paired with a stash.

One blocking item and three suggestions.

---

### 1. `pendingExternalFiles.ts` contains a raw NUL byte, so git treats it as binary

At byte offset 2325 (line 41) the template literal holds a **literal `0x00` byte**, not the `\x00` escape sequence:

```ts
return `${channel}<0x00 byte here>${nodeId}`;
```

```
$ git diff --stat origin/main...HEAD
 .../canvas/application/pendingExternalFiles.ts | Bin 0 -> 3194 bytes

$ file frontend/src/features/canvas/application/pendingExternalFiles.ts
... : data
```

The runtime behavior is fine — a NUL inside a JS string is legal, and using it as a key separator is a good choice precisely because neither a channel name nor a node id can contain one. The problem is entirely at the source level: **this file could not be reviewed in the PR at all.** There is no line diff, no blame, no way to leave an inline comment, and a future rebase or merge touching it cannot be resolved textually. I had to dump the blob and substitute a placeholder for the NUL in order to read it.

Writing it as the escape `\u0000` produces an identical string at runtime and returns the file to text:

```ts
return `${channel}\u0000${nodeId}`;
```

Worth checking whether anything in the repo would have caught this — a `.gitattributes` rule or a lint step that rejects control characters in tracked source would prevent a recurrence, since CI passed cleanly here.

### 2. Drop `file` from the three event payloads so the compiler enforces the new invariant

The contract this PR establishes is "producers must stash **and** publish; consumers read only the stash." Right now that is enforced only by the prose in `ports.ts:175-178`.

I checked the current state: no consumer reads `payload.file` anymore — `useExternalFileHandoff` destructures `{ nodeId: targetId }` and nothing else. So `file: File` on `'upload-node/external-file'`, `'video-node/external-file'` and `'audio-node/external-file'` is now dead weight that still typechecks.

The failure mode that leaves is precisely the bug class this PR fixes: a future producer writes `publish('upload-node/external-file', { nodeId, file })`, forgets `stashExternalFile`, compiles cleanly, and silently does nothing. Removing `file` from those three `CanvasEventMap` entries makes that unrepresentable — the only way to hand a `File` to a node becomes `stashExternalFile`. `spawnExternalAssets.ts:40`, which references the payload type, would fall out of the same change.

### 3. `VideoNode`'s type guard silently swallows an already-consumed file

`VideoNode.tsx:1394`

```ts
if (!isVideoFile(file)) return;
```

Because `take` is consuming, a non-video file reaching this branch is removed from the stash and dropped with no toast and no trace. `AudioNode` and `UploadNode` both delegate validation to `processFile` / `handleMediaFile`, which surface a rejection to the user.

This is currently unreachable — the only stasher on that channel, `UploadNode.morphToVideoWithFile`, already gates on `isVideoFile` before morphing — so it is defensive code. Flagging it because it is the one place where the switch to consuming semantics converts a former no-op into silent data loss.

### 4. Smaller notes

- **Eviction runs before the key-exists check.** `pendingExternalFiles.ts:50` evicts the oldest entry whenever `pending.size >= MAX_PENDING`, including when the incoming key is already present. Re-stashing onto an existing node at exactly 64 entries discards an unrelated pending file for no gain. Moving the check after a `pending.has(key)` test avoids it.

- **`spawnExternalAssetNodes` does not call `focusNewNodeIfLowZoom`.** Unlike the drop and paste entry points, its nodes can stay as shells at low zoom, with their files sitting in the stash until the user zooms in. That is still a strict improvement over the current behavior (eventually correct rather than permanently lost), and skipping the viewport jump is defensible since those nodes are anchored to a node the user is already looking at — worth one line in the comment saying the omission is deliberate, so the next reader does not "fix" it.

- **`onFile` in the effect deps churns the subscription.** All three node call sites memoize it with `useCallback`, but the new test passes an inline arrow (`external-file-handoff-low-zoom.test.tsx:57`), so every render tears down and recreates the subscription. Harmless today — `publish` happens in a rAF so nothing can interleave with the teardown, and a `drain()` against an empty stash is a no-op — but holding `onFile` in a ref would remove the churn on `VideoNode`, which re-renders frequently.

n7s4:
@lywaterman 感谢 review，逐条改了，都在 `2cc6b4c3`。

**1（blocking）裸 NUL 字节** — 已改成 JS 转义写法，`file` 现在报 `UTF-8 text`，`main...HEAD` 的 diff 里 `pendingExternalFiles.ts` 是正常的 79 行文本新增，可以逐行看了。顺手扫了一遍本 PR 全部改动文件，只有这一处。同时在 `keyOf` 上留了注释，说明为什么必须写转义而不是嵌真字节。

关于你提的「CI 全绿地放过了一个二进制源文件，要不要加个 .gitattributes 或 lint 关卡」：我扫了一遍仓库现有被跟踪的文本文件，没有别的控制字符，所以这个关卡是可以加的。但它要带一份二进制白名单（图片/字体/wasm 都得排除），跟本 PR 的画布投递链没有关系，我倾向单开一个 PR 做，免得在这条 bugfix 里动共享的 lint 关卡。

**2 payload 里的 `file`** — 采纳。`upload-node/external-file`、`video-node/external-file`、`audio-node/external-file` 三条现在都只剩 `{ nodeId: string }`，「只 publish 不 stash」这种写法直接编译不过。如你所料 `spawnExternalAssets.ts:40` 的 `deps.publish` 类型跟着收窄了，另外把 `spawn-external-assets.test.ts` 里读 `payload.file` 的两处断言改成对暂存断言（`takeExternalFile(...)` `toBe` 原 File，仍然比身份不比结构），并补了 `beforeEach` 清暂存 —— 那个文件里每个用例的节点 id 都从 `up-0` 重新开始，模块级暂存不清会串味。

**3 VideoNode 静默吞文件** — 采纳。走到 `consumeExternalFile` 时文件已经从暂存里被 take 走了，直接 return 等于丢在地上。现在打一句 `[video-node] external file "..." (...) is not a video; dropped`，口径对齐 `[upload-node]` / `[spawn-external-assets]`。

**4b `focusNewNodeIfLowZoom` 的缺省** — 是有意的，已在 `spawnExternalAssetNodes` 的文档注释里写明：拉到 0.6 是给「从画布空白处凭空新建节点」用的（落点用户未必看得见），而这个函数是从一个已选中的目标节点派生上游素材，视野本来就在目标节点上，再改一次缩放反而抢镜头。

**4c `onFile` 进 ref** — 采纳，订阅现在只依赖 `[channel, nodeId]`。理由按你说的写进注释了：调用方忘了 `useCallback` 时，退订/重订之间会留出丢事件的窗口，「挂载补投」也会每次渲染都跑一遍。`useRef(onFile)` 自带首帧值，赋值 effect 声明在订阅 effect 之前，所以挂载那次补投拿到的就是最新回调。

验证：`tsc -b` + `pnpm build` 干净，`src/__tests__/features/canvas` + `src/__tests__/stores` 105 files / 732 tests 全绿，PR 上 16 个 check 全 pass。

lywaterman:
Verified against `2cc6b4c3`. All six items are addressed, including 4a, which your reply did not mention but the code does fix (`pending.size >= MAX_PENDING && !pending.has(key)`).

- **1** — `file(1)` now reports UTF-8 text, the blob has zero control characters, and `main...HEAD` renders the module as a normal 79-line text addition. The comment on `keyOf` explaining why the escape is mandatory is the right thing to leave behind.
- **2** — all three payloads are down to `{ nodeId: string }` and every publish site narrowed with them. "Publish without stashing" is now a compile error rather than a convention.
- **3** — the warning matches the `[upload-node]` phrasing, so a dropped file leaves a trace.
- **4b** — agreed, and the reasoning now lives in the docstring, which is what I was after: the omission reads as deliberate instead of overlooked.

On **4c** I checked the ordering rather than the diff, since latest-ref is easy to get subtly wrong: `useRef(onFile)` carries the first render's value, the assignment effect is declared before the subscribe effect so it runs first on mount and on any `nodeId` change, and updates re-run only the assignment effect. `drain` closes over `channel`/`nodeId` — both in the dependency array — and reads the callback through `onFileRef.current` at call time, so there is no stale closure. Correct as written.

The test changes hold up too: moving the assertions from `payload.file` to `takeExternalFile(...)` keeps them on identity (`toBe`) rather than structure, and the `beforeEach`/`afterEach` reset is genuinely required — every case in that file restarts node ids at `up-0`, so a module-level stash would leak across cases.

Agreed on deferring the control-character guard to its own PR. It needs a binary allowlist for images, fonts and wasm, and that is a shared lint gate — not something to land inside a canvas bugfix.

Nothing outstanding from me.

lywaterman (approved):

lywaterman (commented):
The design is right, and one part deserves calling out specifically: **removing `file` from the event payload type** so that "publish without stash" cannot compile. That is a much stronger guarantee than a comment telling future authors to remember — the constraint now lives in the type system, and a new delivery site physically cannot get it wrong. The consumptive `take` also removes the need for a dedup flag, which is a nice reduction.

One finding.

## `MAX_PENDING = 64` eviction reintroduces the exact bug this PR fixes

```js
if (pending.size >= MAX_PENDING && !pending.has(key)) {
  const oldest = pending.keys().next();
  if (!oldest.done) pending.delete(oldest.value);
}
```

I checked `Canvas.tsx` and `spawnExternalAssets.ts`: **there is no file-count limit anywhere on the drop path.** And the drop handler is a synchronous `forEach` — every `addNode` + `stashExternalFile` runs to completion before React mounts any of the new nodes.

So dropping 100 images in one gesture: eviction starts at the 65th stash, and **the first 36 nodes are created but stay permanently empty** — the exact symptom this PR fixes, silently.

The comment says "正常用法下 map 里同时不会有几条", but that premise doesn't hold for bulk import, and bulk-importing reference material is a mainline workflow for a canvas tool.

What makes this worth fixing rather than noting: it contradicts the rule this PR itself establishes. In `VideoNode.tsx` you added:

```js
// 走到这里文件已经从暂存里被取走了,直接 return 等于把它丢在地上 ——
// 留一句警告,别静默。
```

The eviction path is exactly "dropping it on the floor", and it doesn't even warn.

Two reasonable fixes:
- `console.warn` on eviction, matching the `[video-node]` precedent you just set; or
- raise the cap well above any plausible single drop. The leak this cap defends against comes from *nodes deleted before consumption*, which is a low-frequency event — 64 is simultaneously too tight for bulk drops and too loose to bound that leak promptly.

Worth adding a 65-file test case either way; the current suite has no coverage at the cap boundary.

PR #308, opened 2026-08-17
fix: synchronize custom model configuration

### LINKED ISSUE #291: feat: 优化自定义模型配置并引入开放的媒体协议适配层
## 前置检查 / Pre-flight checks

- [x] 我已搜索过现有 issue 与 Discussions，确认这不是重复 / I searched existing issues and Discussions

## 要解决的问题 / Problem

DramaClaw CE 的自定义媒体模型主要通过 New API 类网关调用。图片、视频供应商在请求字段、素材角色、鉴权、文件上传、异步任务、状态查询、结果解析和错误格式方面存在差异。如果这些差异直接进入 DramaClaw，新增或调整供应商时就需要修改创作业务代码并同步发布 DramaClaw。

本方案引入一个面向 DramaClaw CE 的轻量开源模型适配层，让 DramaClaw 使用稳定的公共媒体协议，供应商差异集中在网关适配器中。

## 项目定位 / Project positioning

新仓库：[`dramaclaw/dramaclaw-gateway`](https://github.com/dramaclaw/dramaclaw-gateway)

产品名称：**RelayClaw CE**

RelayClaw CE 基于 New API 官方代码 fork 并持续同步上游。它不是现有商业虾驿功能的开源复制版，也不追求同步商业虾驿中的计费表达式、调用审计、结果归档、运营后台、价格展示或其他增强功能。

RelayClaw CE 只做必要的轻量改动：

1. 保留 New API 官方已有的渠道、模型路由、令牌和基础管理能力。
2. 增加 DramaClaw 所需的 `DC-Media-Protocol` 请求规范化、校验和响应归一化。
3. 将统一图片、视频请求转换为各供应商上游协议。
4. 处理媒体上传、异步任务提交、查询、取消、结果提取和错误转换。
5. 通过协议契约测试保证 DramaClaw 与 RelayClaw CE 的兼容性。

## 架构边界 / Architecture

```text
DramaClaw CE 创作业务 / 虾画节点
              |
              | DC-Media-Protocol v1
              v
RelayClaw CE（dramaclaw-gateway）
  |- New API 官方已有渠道
  |- DC-Media-Protocol 规范化与校验
  `- 必要的供应商 Channel Adapter
              |
              v
        图片 / 视频模型供应商
```

职责划分：

- **DramaClaw 模型目录**：声明模型身份、网关模型名、支持模式、分辨率、比例、时长和素材数量限制。
- **DramaClaw**：根据用户选择校验模型能力，构造规范请求，不发送供应商专属字段。
- **RelayClaw CE**：校验公共字段组合，根据素材结构选择调用形态，并转换为供应商协议。
- **供应商适配器**：负责字段映射、鉴权、上传、轮询和响应转换，不得静默丢弃素材或改变素材角色。

## DC-Media-Protocol v1

协议草案位于：

```text
dramaclaw-gateway/dc-media-protocol.md
```

第一阶段覆盖：

- 图片文生图；
- 图片编辑和多图参考；
- 视频文生视频；
- 单图图生视频；
- 首帧和首尾帧视频；
- 多图参考和图片/视频/音频全能参考；
- 视频编辑；
- 异步任务提交、查询和取消；
- 统一结果、状态、错误码和上游 request ID。

文本、Embedding、Rerank 和现有音频兼容接口继续使用 New API 官方已有协议。除非后续出现明确需求，不在 `DC-Media-Protocol v1` 中重复设计。

### 公共接口

```http
POST /v1/images/generations
POST /v1/images/edits
POST /v1/video/generations
GET  /v1/videos/{task_id}
DELETE /v1/videos/{task_id}
```

具体路径可以复用 New API 当前路由，但同一协议版本下请求和响应语义必须稳定。

### 不增加 mode 字段

`DC-Media-Protocol v1` 不增加顶层 `mode`。DramaClaw 的业务模式在发送到网关前转换为具有明确角色的素材字段，RelayClaw CE 根据素材组合选择供应商调用形态。

业务模式之间允许以下包含和归一化关系：

```text
图生视频 ⊂ 图片参考 ⊂ 全能参考
```

因此：

- 单图图片参考或单图全能参考可以归一化为图生视频；
- 多图请求归一化为图片参考；
- 存在视频或音频素材时归一化为全能参考；
- 网关不需要恢复 DramaClaw 原始界面模式，只需正确转换素材和参数。

### 调用形态推断优先级

网关先校验互斥字段，再按以下顺序识别：

| 优先级 | 请求特征 | 调用形态 |
|---:|---|---|
| 1 | `duration = "auto"`、`metadata.ratio = "auto"` 且存在 `reference_videos` | 视频编辑 |
| 2 | 存在 `metadata.last_frame_image`，可同时存在顶层 `image` | 首尾帧或尾帧视频 |
| 3 | 存在顶层 `image` | 首帧视频 |
| 4 | 存在 `metadata.reference_videos` 或 `metadata.reference_audios` | 全能参考 |
| 5 | `metadata.reference_images` 数量大于 1 | 图片参考 |
| 6 | `metadata.reference_images` 数量等于 1 | 图生视频 |
| 7 | 不存在任何输入素材 | 文生视频 |

约束：

- 顶层 `image` 只表示视频首帧；参考图只能进入 `metadata.reference_images`。
- 顶层 `image` 与 `metadata.reference_images` 不得同时出现。
- `last_frame_image` 不得与参考图片、视频或音频混用。
- 第一张参考图不得自动提升为首帧。
- 供应商不支持多图或某类素材时必须返回明确错误，不得静默忽略。
- 新请求统一使用 `auto`、小写分辨率和 `duration`；`adaptive`、`seconds`、`size` 等只在兼容边界读取。

## 适配器策略 / Adapter strategy

优先复用 New API 官方已有适配器。只有现有适配器无法正确处理 `DC-Media-Protocol` 时，才增加或扩展独立 Channel Adapter。

复杂供应商能力包括：

- multipart 或 Base64 图片编辑；
- 临时素材上传；
- 签名鉴权；
- 异步任务轮询；
- 真人素材审核；
- 多阶段任务创建；
- ComfyUI Workflow 路由和输入注入。

每个新增适配器必须：

- 只读取规范字段和明确的兼容字段；
- 保留显式 `false`、`0` 等可选参数值；
- 不根据模型名称猜测公共业务能力；
- 不静默覆盖比例、分辨率、时长或素材；
- 返回统一任务状态和可读错误；
- 提供对应的协议转换测试。

## 非目标 / Non-goals

RelayClaw CE 第一阶段不建设或迁移：

- 商业虾驿的动态计费表达式和价格展示；
- 调用审计和管理端排障增强；
- 图片、视频和音频结果归档；
- 充值、订阅、支付和商业运营能力增强；
- 商业虾驿的渠道元数据和前端定制；
- 与协议转换无关的虾驿历史修改。

New API 官方原本存在但 DramaClaw 不使用的功能可以暂时保留，不为精简体积而大范围删除，避免增加上游同步成本。

## 仓库与上游同步 / Repository strategy

```text
origin   -> dramaclaw/dramaclaw-gateway
upstream -> QuantumNous/new-api
```

维护原则：

1. `main` 保持可运行，并定期同步 New API 官方稳定节点。
2. 协议层和供应商适配器使用独立、可追踪的提交。
3. 不整体合入商业虾驿仓库，只按实际需要移植协议转换代码。
4. 每次上游同步使用独立 PR，先验证 New API 原测试，再验证 DC 协议契约。
5. 保留 New API 的许可证、NOTICE、版权、来源说明和必要署名。

### 阶段 1：协议定稿

- 完成 `dc-media-protocol.md`。
- 固定图片、视频请求结构和素材互斥规则。
- 固定调用形态推断优先级。
- 固定任务状态、结果和错误结构。
- 在 DramaClaw 与 RelayClaw CE 建立相同请求样例的契约测试。

### 阶段 2：最小可用链路

- 图片文生图和图片编辑。
- 视频文生视频、单图图生视频和首帧视频。
- 异步任务创建与查询。
- 使用至少一个真实供应商完成 Dramaclaw CE 端到端验证。

### 阶段 3：完整媒体能力

- 首尾帧。
- 多图参考。
- 图片、视频和音频全能参考。
- 视频编辑。
- ComfyUI 本地 Workflow。
- 任务取消和更多供应商适配。

### 阶段 4：扩展文档

- 适配器开发指南。
- 示例供应商适配器。
- 新模型接入检查清单。
- 上游同步和发布说明。

## 验收标准 / Acceptance criteria

- DramaClaw CE 只发送 `DC-Media-Protocol` 规范字段，不包含供应商专属字段。
- 新增一个使用已有协议的模型时，不需要修改 DramaClaw 业务代码。
- 新增特殊供应商时，改动集中在 RelayClaw CE 独立适配器。
- 单图、多图、视频/音频参考和视频编辑按照固定优先级选择调用形态。
- 首帧、尾帧和参考素材不会相互提升或混用。
- `ratio`、`resolution`、`duration`、`quality`、`watermark` 和参考素材具有明确类型与优先级。
- 不支持的字段组合在调用上游前返回稳定错误。
- 异步任务经过不同供应商后返回一致状态和结果结构。
- 协议契约测试覆盖正常请求、边界值、兼容输入和非法组合。
- RelayClaw CE 可独立发布；供应商协议变化不要求同步发布 DramaClaw。

## 兼容与合规 / Compatibility and compliance

- 保留 New API 原有 OpenAI 兼容接口，避免破坏官方已有渠道。
- 旧字段只在明确的兼容边界归一化，新代码不得继续产生旧字段。
- 删除兼容逻辑前必须确认历史项目、队列任务和现有 Dramaclaw 版本不再依赖。
- RelayClaw CE 遵循 New API 的 AGPLv3 许可要求，保留许可证、NOTICE、版权及源码提供方式。

## 备选方案 / Alternatives

不采用以下方案：

- 在 DramaClaw 内直接维护各供应商 Provider Adapter：会让供应商变化持续侵入创作业务。
- 将商业虾驿整体开源或同步到新仓：会引入与 RelayClaw CE 目标无关的计费、审计、归档和运营复杂度。
- 完全使用任意 URL 和 JSON 模板映射：无法可靠覆盖异步任务、上传、签名、审核和 ComfyUI Workflow，也容易产生不可验证的协议分支。

最终边界保持为：**DramaClaw 负责创作业务和模型能力声明，RelayClaw CE 负责轻量、稳定、可扩展的供应商协议转换。**

NewYuee:
## What changed

- keep the quick custom-model JSON synchronized with channels and models saved in Advanced Settings
- hydrate browser model settings from the backend, including authoritative empty embedding and media-model snapshots
- expose the MiniMax H3 workflow template only inside a configured ComfyUI channel and bind it to `MiniMax-H3-local`
- refresh the Chinese and English custom model configuration guides

## Why

The quick JSON editor and Advanced Settings could drift apart. Stale browser state could restore models already removed from the backend, and loading the H3 template into an existing ComfyUI channel could retain an unrelated model name.

## User impact

Custom model configurations now reflect the saved backend state. Advanced edits automatically become the current custom profile, and the ComfyUI template shortcut produces a consistent MiniMax H3 model configuration.

## Validation

- `cd frontend && pnpm build`
- `git diff --check origin/main..HEAD`
- manually cleared the isolated browser model-profile storage and opened Settings -> Custom without a runtime exception

## Notes

The full frontend test suite was also run locally under Node 26: 2272 tests passed and 6 existing environment-sensitive tests failed around experimental `localStorage` and `FormData` behavior. Repository CI uses Node 22.

Closes #291


lywaterman (commented):
## Code review — custom model configuration sync

Scope reviewed: `frontend/src/components/settings/settings-dialog.tsx`, `frontend/src/stores/settingsStore.ts`, and both updated docs (docs match the code; no findings there).

The core issue: the new "backend snapshot → store" sync effects don't distinguish between writes made by someone else and writes they just made themselves, so a successful save gets rolled back by the effect it triggers.

### High

**1. `settings-dialog.tsx:2756` — save can wipe the embedding model config**

`handleSave` (line 2863) calls `setEmbeddingModel(saved)`. That mutates `localSavedEmbeddingModel` / `localSavedKey`, which are deps of the sync effect, so the effect re-runs immediately — but `savedEmbeddingModel` still holds the **pre-save** snapshot from React Query, because the `invalidateQueries` refetch has not resolved yet.

With no embedding model previously configured, the backend snapshot is `{}`, so `backendSnapshotLoaded === true`, `next === undefined`, `nextKey === "null"`, and the guard passes → `setEmbeddingModel(undefined)` + `setLocalModel(undefined)`. The form blanks out and localStorage loses the entry immediately after "saved". If the refetch then fails, or the user closes the dialog in that window, the configuration is gone.

Suggested fix: have the effect ignore its own writes (e.g. track the last snapshot key it wrote in a ref) rather than re-deriving from `savedEmbeddingModel` on every store change.

### Medium

**2. `settings-dialog.tsx:3178` — same race in `MediaModelsBlock`**

`handleSave` (line 3354) calls `setMediaModels(saved)` → `localSavedMediaModelsKey` changes → the effect re-runs against the stale `savedMediaModelsKey` → `setMediaModels(oldFromBackend)` writes the pre-save snapshot back into the persisted store, and `setLocalMediaModels` reverts the list. It self-heals once the config refetch lands, but any newly added media model visibly disappears and reappears, and the stale set is persisted in between.

**3. `settings-dialog.tsx:1827` — profile selection silently flips to "custom"**

`setSelectedProfileKind("custom")` + `saveStoredQuickProfiles({ selected: "custom", ... })` now fire on *any* `featureModelConfig` change, not just user edits. Combined with findings 1 and 2, simply opening Settings → Custom mutates the store during backend hydration, the sync effect runs, and the user's "Recommended" selection is silently flipped to "My Config" and persisted to localStorage.

Repro: apply Recommended, then reopen Settings → Custom with any backend media model the profile doesn't list (e.g. a ComfyUI model, which `handleApply` deliberately preserves at lines 1924-1929) → `syncedJson !== appliedProfileJson` → the radio flips without the user touching Advanced Settings. The previous `selected: selectedProfileKind` preserved the choice.

**4. `settings-dialog.tsx:4574` — template button overwrites an existing ComfyUI model name**

`model_name: defaultComfyWorkflows.model` is now unconditional; the old handler used `currentModel || defaultComfyWorkflows.model`.

A user with an existing ComfyUI channel named e.g. `my-wan-video` who clicks "Load MiniMax H3 Template" to restore a missing workflow — the exact use the updated docs advertise at `docs/en/getting-started/configuring-models.md:229` — gets `model_name` silently rewritten to `MiniMax-H3-local`. `MediaModelsBlock`'s managed-draft effect (lines 3190-3203) then deletes their `my-wan-video` media-model entry (`COMFY_WORKFLOW_MANAGED_CONFIG_KEY === true` and no longer in `comfyWorkflowModels`) along with its label and config, and adds a fresh `MiniMax-H3-local` row. Workflows are merged, but the model binding and its media-model config are lost with no warning.

**5. `settings-dialog.tsx:4721` — ComfyUI section can become a dead end**

Moving "Load MiniMax H3 Template" inside `ComfyUIWorkflowsEditor` makes it reachable only when a ComfyUI channel row already exists; the removed handler called `addFeatureProviderChannel("comfyui")` itself.

But `comfyui` is **not** in `FEATURE_MODEL_PROVIDERS` (`settingsStore.ts:28-81`). When the NewAPI channel-types query is unavailable — `channelTypesEnabled` false, or `/channel-types` returning 502 (`model_gateway.py:927`) — `providerOptions` falls back to `FEATURE_MODEL_PROVIDERS.filter(p => allowedProviderSet.has(p))` (line 4227), which is empty in `comfyOnly` mode. `availableProviders` is empty, `handleAdd` returns early (line 4260), and the Hybrid-mode "ComfyUI Configuration" section becomes a dead end: no channel can be added, so the template is unreachable. Adding `"comfyui"` to the fallback list would fix it.

### Low

**6. `settings-dialog.tsx:4556` — dropped base URL self-heal**

The removed `useEffect` that auto-filled `http://127.0.0.1:8188` when a ComfyUI channel had template workflows but an empty base URL is gone; only the button click fills it now. A ComfyUI channel rehydrated from persisted state, or synced from the backend with an empty `baseUrl`, will now fail the save with `baseUrl is required for provider comfyui` (`model_gateway_settings.py:474`) instead of self-healing. Recoverable by typing the URL, but it's a silently dropped guard.

**7. `settings-dialog.tsx:1645` — `fal_ai` channel creation can fail with `unknown provider`**

`ensureChannel` only re-attaches a channel `type` for `comfyui` (`...(provider === "comfyui" ? { type: 63 } : {})`). The recommended profile now introduces `fal_ai` with `type: 61`, and `fal_ai` has **no** entry in the backend `PROVIDER_PRESETS` (`newapi_provisioner.py:26-140`).

Whenever the sync has to *create* the `fal_ai` channel (applied profile predates this PR, or is hand-written without it), the regenerated custom JSON carries no `type`, and `build_channel_payload` raises `unknown provider; pass type when using a custom provider` (`newapi_provisioner.py:767`) unless a previously-saved type happens to be available as a fallback. The `comfyui` special case shows the hazard was already known; `fal_ai` needs the same treatment.

**8. `settings-dialog.tsx:3162` — normalization mismatch causes a redundant write on every mount**

In `MediaModelsBlock`, `fromBackend` keeps `upstreamModel` / `label` untrimmed, while `setMediaModels` (`settingsStore.ts:556-571`) trims them and drops whitespace-only labels. Any backend value with surrounding whitespace makes `localSavedMediaModelsKey !== nextKey` permanently true, so `setMediaModels` re-fires on every mount and rewrites localStorage. It converges after two effect passes rather than looping, but the two shapes should be produced by one shared normalizer. Note `settingsStore.ts:586-593` fixes exactly this class of mismatch for `batchSize` on the embedding side — the media side wasn't given the same treatment.

### Checked and cleared

- `savedEmbeddingModel` keeping `config?.provisioner?.embeddingModel` while `savedMediaModels` got the `config ? … : undefined` guard is safe: `build_provisioner_status` (`newapi_provisioner.py:1370`) always emits both keys, and `get_newapi_embedding_model_config` returns `{}` (not omitted) when unconfigured, so `!== undefined` correctly distinguishes "loading" from "loaded empty".
- `updateFeatureProviderChannel` merges partials, so dropping `priority: current?.priority ?? 0` from the template handler does not reset priority.
- `addFeatureProviderChannel` / `updateFeatureProviderChannel` in `FeatureModelsBlock` remain used by the effect at line 2417, so removing the old handler leaves no unused bindings.
- The embedding effect's `setEmbeddingModel` write does not loop: the `batchSize` addition in `settingsStore.ts` makes the store shape and the `fromBackend` shape stringify-identical.

NewYuee:
已按本轮 8 项检视意见完成修复，更新提交：`a295a66d`。

1. Embedding 同步只在后端快照 key 变化时执行，不再因保存后的本地 store 写入而使用旧快照回滚。
2. Media Models 使用同样的后端快照同步边界，新增或编辑模型保存后不会再消失、回闪。
3. 为模型配置 mutation 增加 `user` / `hydrate` / `profile` 来源，并只用用户修改 revision 触发自动切换到 Custom；后端 hydration 和应用预设不再静默改变 profile。
4. 载入 MiniMax H3 模板前若 ComfyUI 已绑定其他模型，会明确提示模型切换及影响，用户确认后才执行，不再静默覆盖。
5. `/channel-types` 不可用时，provider fallback 会合并当前区块允许的 provider，Hybrid 的 ComfyUI 渠道仍可创建。
6. 恢复 ComfyUI 模板 workflow 存在但 `baseUrl` 为空时自动补全 `http://127.0.0.1:8188`，并兼容旧数据缺字段的情况。
7. 自定义 provider 类型统一兜底：`fal_ai=61`、`comfyui=63`，profile 同步和渠道保存都不会丢失 type。
8. 后端媒体模型快照和 store 共用 `normalizeMediaModelEntries`，统一 trim、空 label、默认值规则，避免每次挂载重复写入。

补充了 store mutation 来源和媒体模型归一化回归测试，并更新 license inventory。

验证结果：

- `pnpm build`：通过
- `pnpm test src/__tests__/stores/settings-store-feature-model-config.test.ts src/__tests__/i18n/locales-json.test.ts`：10/10 通过
- `.venv/bin/pytest tests/test_p0b_compliance_generator.py::test_license_inventory_covers_current_git_index`：通过
- 完整 `pnpm test`：2274/2280 通过；其余 6 项为当前 Node 26 环境的 localStorage 与 FormData/MSW 兼容失败，失败文件不在本次改动范围内

lywaterman (commented):
## Follow-up review — `a295a66d`

Reviewed `a295a66d` against its parent `b3eaacd3`, read the post-fix file in full context, ran the new test (2/2 pass) and `tsc -b` (clean) in a throwaway worktree.

### Verdicts on the 8 original findings

| # | Verdict | Mechanism |
|---|---------|-----------|
| 1 | **Fixed** | `EmbeddingModelBlock` keys hydration on a `lastSyncedBackendKey` ref (`settings-dialog.tsx:2758`) and drops `localSavedKey` / `localSavedEmbeddingModel` from the deps, so a local save can no longer re-enter the effect. It also compares against a fresh `useSettingsStore.getState()` read instead of the render-time store value. |
| 2 | **Fixed** | Identical mechanism in `MediaModelsBlock` (`:3142`, `:3191-3220`). |
| 3 | **Mostly fixed** | New `featureModelConfigUserRevision` counter, incremented only when `source` is absent or `'user'` (`nextFeatureModelConfigUserRevision`, `settingsStore.ts:456`). All hydration sites pass `{source:'hydrate'}`; `handleApply` passes `{source:'profile'}`. I enumerated every call site of the 8 mutators across `frontend/src` — the tagging is complete and correct. Two residual holes remain, N3 and N4 below. |
| 4 | **Partially fixed** | A `confirmDialog` gate was added (`:4617`) when `model_name` differs, but the destructive write itself is unchanged — and the handler now reads pre-`await` state (N2). |
| 5 | **Fixed for `comfyui`** | The offline fallback unions `allowedProviders` into the option list (`:4259`), so the ComfyUI-only section is no longer a dead end. Still broken for `fal_ai` (N5). |
| 6 | **Fixed** | Self-heal restored, scoped to the ComfyUI row and tagged `hydrate` (`:4510`). Restored verbatim, including a pre-existing flaw (N1). |
| 7 | **Fixed** | `CUSTOM_PROVIDER_CHANNEL_TYPES = {fal_ai: 61, comfyui: 63}` (`:1379`), consistent with the recommended profile's `type: 61` (`:1467`) and the backend's hardcoded `63` (`model_gateway.py:397`). Applied at `ensureChannel` plus both the `buildChannels` and `handleSaveChannels` sites. |
| 8 | **Fixed** | Shared exported `normalizeMediaModelEntries` (`settingsStore.ts:440`) is applied to `fromBackend` *and* is what `setMediaModels` uses, so key order, trimming and defaults match and the comparison converges. Bonus hardening: the `typeof upstreamModel === 'string'` guard removes a `.trim()`-of-`undefined` crash on a backend entry missing `upstreamModel`. |

### New findings

**N1. `settings-dialog.tsx:4510` — medium — the ComfyUI Base URL field can no longer be cleared**

The restored self-heal fires on *any* transition to empty/whitespace while template workflow IDs are present, and the input at `:4727` is fully controlled off `channel?.baseUrl`. Select-all + Backspace → store `baseUrl: ""` → the effect immediately writes back `http://127.0.0.1:8188` → the field snaps back. Needs a "already self-healed once" ref, or to run only on mount / on template load rather than on every empty-baseUrl render.

**N2. `settings-dialog.tsx:4641` — medium — stale state written back after the confirm dialog**

`handleLoadDefaultComfyWorkflows` became `async`, but `currentSettings` (`:4615`) and `channel?.baseUrl` (`:4642`) are captured **before** the `await confirmDialog(...)`. `useModelGatewayConfig` has no `staleTime`, so React Query's refetch-on-window-focus can land while the modal is open; `FeatureModelsBlock`'s hydration effect (`:2441`) then rewrites `providerChannels` from the backend, and on confirm this handler writes the stale pre-await `currentSettings` back, silently discarding the freshly hydrated channel settings. The file already uses the correct pattern two functions up (`useSettingsStore.getState()` at `:4592`) — re-read the channel after the await.

**N3. `settings-dialog.tsx:1800` — medium — edits made in hybrid mode can be permanently lost from the profile JSON**

`previousAdvancedSettingsUserRevision` is seeded from the live revision at mount, and the revision is deliberately excluded from persistence (`settingsStore.ts:884`). `QuickLocalNewApiSetup` only renders when `mode === "custom"` (`:494`), while the hybrid branch renders a fully editable ComfyUI `FeatureModelsBlock` (`:508`).

Scenario: the user switches to hybrid, edits the ComfyUI channel (revision++, nobody watching), then switches back to custom → the component remounts with `ref === current revision`, so the sync never fires and the stored/displayed profile JSON permanently omits those edits until some *unrelated* later edit bumps the revision again. Before this fix, the `useRef("")` seed guaranteed a first-run reconciliation. Seeding from a store snapshot captured at the last successful sync — or persisting the revision — would close this.

**N4. `settings-dialog.tsx:1804` — low — residual of finding #3 on first run**

The `!appliedProfileJson` recovery branch is not gated by the revision at all. A user who has never applied a profile but whose backend already has channels (provisioned elsewhere, or by CLI) will have hydration alone flip `selectedProfileKind` to `"custom"` and overwrite `customProfileJson` — the exact symptom #3 described, narrowed to first run.

**N5. `settings-dialog.tsx:4259` — low — `fal_ai` still hits the dead end from #5**

The offline fallback list is `FEATURE_MODEL_PROVIDERS ∪ allowedProviders`, but `FEATURE_MODEL_PROVIDERS` (`settingsStore.ts:28-82`) contains neither `comfyui` nor `fal_ai`, and the non-ComfyUI sections pass `allowedProviders: undefined`. So when `/channel-types` is unavailable the user still cannot add the `fal_ai` channel the recommended profile depends on — the same dead end as #5, for the other key of the `CUSTOM_PROVIDER_CHANNEL_TYPES` map introduced 2900 lines above. Adding `...Object.keys(CUSTOM_PROVIDER_CHANNEL_TYPES)` to that union fixes both uniformly.

### Notes (not filed as findings)

- `handleSave` at `:3387` puts the **un-normalized** backend response into local state while `:3386` normalizes it into the store — same class as #8, but it converges on the next refetch, so it's cosmetic.
- The template merge at `:4648` is `{...template, ...existing}`, so existing workflows win. After confirming "switch to MiniMax H3", a channel whose workflows already use the template IDs gets only `model_name` rewritten. Reaching an actually inconsistent state requires a contrived setup.
- Translation keys verified present and correctly nested under `settings.modelConfig.featureModels` in both `en` and `zh`; `confirmDialog` never rejects, so `void onLoadDefaultTemplate()` at `:5015` is safe.

NewYuee:
已处理 follow-up review 中的 N1、N2、N3、N5，更新提交：`a9cadc05`。

- **N1**：ComfyUI Base URL 自愈现在记录本次挂载是否已执行，并区分用户输入。已有模板且后端初始值为空时仍会自动补全；用户主动清空后不会再次被 effect 写回。
- **N2**：MiniMax H3 切换确认完成后，重新通过 `useSettingsStore.getState()` 获取最新渠道、Base URL 和 settings，再执行模板合并，避免写回 `await` 前捕获的旧状态。
- **N3**：用 store 中的 `featureModelConfigProfileSyncedRevision` 记录 profile JSON 最后同步到哪个用户 revision；该游标与 user revision 一并持久化。Hybrid 模式的用户修改即使发生在 `QuickLocalNewApiSetup` 未挂载期间，切回 Custom 或刷新后也会触发一次同步；`hydrate` / `profile` 写入仍不会推进 user revision。
- **N5**：`/channel-types` 不可用时，fallback provider 列表统一加入 `CUSTOM_PROVIDER_CHANNEL_TYPES` 的全部 provider，因此 `fal_ai` 和 `comfyui` 都可以手动新增。

**N4 未修改**：当没有已应用 profile、但后端已经存在渠道或模型配置时，将其恢复为 Custom JSON 并显示 Custom 是预期行为。此时显示 Recommended 会与实际正在使用的后端配置不一致。

验证：

- `pnpm build`：通过
- `pnpm test src/__tests__/stores/settings-store-feature-model-config.test.ts src/__tests__/i18n/locales-json.test.ts`：10/10 通过
- `pnpm test src/__tests__/lib/queries/model-gateway.test.tsx src/__tests__/stores/settings-store-feature-model-config.test.ts`：5/5 通过
- `.venv/bin/pytest tests/test_p0b_compliance_generator.py::test_license_inventory_covers_current_git_index`：通过

lywaterman (commented):
## Follow-up review 2 — `a9cadc05`

Reviewed `a9cadc05` against the four accepted findings plus the declined one, reading the surrounding code in `settings-dialog.tsx` and `settingsStore.ts` rather than just the hunks.

> **Blocking:** findings 1 and 2 below are new, introduced by this commit, and one of them can overwrite a live backend ComfyUI configuration with empty values. Please resolve both before merge.

### Verdicts

| # | Verdict | Notes |
|---|---------|-------|
| N1 | **Partially fixed** | `comfyBaseUrlEditedByUser` / `didSelfHealComfyBaseUrl` are `useRef`s, so both flags reset on every mount of `ProviderChannelRow`. `QuickLocalNewApiSetup` / `FeatureModelsBlock` are conditionally rendered by mode (`settings-dialog.tsx:486-501`) and the dialog unmounts on close, so clearing the field and then switching tabs or reopening Settings re-runs the self-heal and refills `http://127.0.0.1:8188`. Correct within a mount; still unclearable across mounts. See finding 6. |
| N2 | **Fixed** | `handleLoadDefaultComfyWorkflows` re-reads `useSettingsStore.getState().featureModelConfig.providerChannels[provider]` after the confirm and uses it for both `settings` and `baseUrl` (`:4658-4671`). `provider` comes from `Object.keys(providerChannels)` so it is already store-normalized and the lookup is sound. |
| N3 | **Not fixed — the persisted cursor makes it worse** | See findings 1 and 2. |
| N4 | **I disagree with the decline** | See finding 4. |
| N5 | **Fixed** | `CUSTOM_PROVIDER_CHANNEL_TYPES` contains `fal_ai: 61` (`:1379`), the fallback unions its keys (`:4272`), and `handleSaveChannels` falls back to `customProviderChannelType(provider)` for the type (`:4339`), so the add → save path works end to end with `/channel-types` down. |

### Findings

**1. `settings-dialog.tsx:1833` — high — the persisted cursor is consumed against a store that has no channels yet**

`prepareFeatureModelSettingsForPersistence` (`settingsStore.ts:473`) deliberately blanks `providerChannels` / `providerKeys` while keeping `featureModels` / `mediaModels` / `embeddingModel`.

Scenario: the user edits the ComfyUI channel config in hybrid mode (Quick panel unmounted, `userRevision` → N+1, `syncedRevision` stays N), restarts the app, and opens Settings in custom mode. `QuickLocalNewApiSetup` mounts and its effect fires in the first commit — before the async `savedProviderChannels` fetch has hydrated channels — so `advancedSettings.providerChannels` is `{}`. `syncQuickProfileFromAdvancedSettings` therefore never applies the hybrid channel edits (its `for (const channel of Object.values(settings.providerChannels))` loop is empty), yet `markAdvancedSettingsProfileSynced()` advances the cursor to N+1. When hydration lands it uses `{source:"hydrate"}`, which doesn't bump the revision, so the sync never re-runs. The very edits the cursor was added to preserve are consumed and dropped.

**2. `settings-dialog.tsx:1844` — high — same path, but it can overwrite the live backend with empties**

Worse than a no-op. With `providerChannels` empty but `mediaModels` persisted, `ensureChannel(entry.provider)` synthesizes the missing channels with `baseUrl: ""` and `settings: normalizeProviderChannelSettings(provider, {})` — which returns `{}` for comfyui (`:4801-4814`). For a hybrid user with ComfyUI media models, the newly written `customProfileJson` / `appliedProfileJson` now contain a `comfyui` channel with an empty baseUrl and empty settings. `comfyui` is exempt from the `missingKeyChannels` guard (`:1920`), so a later Apply sends it through `saveProviderChannels` and overwrites the real backend ComfyUI baseUrl and workflow settings with empties.

**3. `settingsStore.ts:596` — medium — the cursor records the wrong revision**

`markFeatureModelConfigProfileSynced` records `state.featureModelConfigUserRevision` (the store's *latest*) rather than the revision the sync actually consumed. Passive effects flush lazily, and React flushes them at the start of the next scheduled render — by which point zustand already holds the newer value. Two user edits landing before the first effect flush means the effect computes `syncedJson` from the rev-N+1 closure but marks synced = N+2; the N+2 render then short-circuits at `:1833` and the last edit never reaches the profile JSON. Pass the consumed revision as an argument instead.

**4. `settings-dialog.tsx:1806` — medium — why I reject the N4 decline**

The `!appliedProfileJson` recovery branch fires against the same half-hydrated store: `hasExistingAdvancedSettings` is satisfied by the *persisted* `featureModels` / `mediaModels` / `embeddingModel` alone, while `providerChannels` is `{}`. It then writes `recoveredJson` as both `customProfileJson` and `appliedProfileJson`, marks it synced, and never revisits it (later backend hydration is `source:"hydrate"`, no bump).

So the profile presented as "Custom, applied" is built from `RECOMMENDED_LOCAL_NEWAPI_PROFILE` channels plus synthesized empty-baseUrl channels — it does not describe the live backend config. The decline rests on "showing Recommended would misrepresent the live backend config", but the recovered Custom JSON misrepresents it too, and unlike a label it is persisted and applyable (finding 2). Either gate recovery on channels actually being hydrated, or keep the label without claiming `appliedProfileJson`.

**5. `settingsStore.ts:894` — medium — root cause of finding 1**

`partialize` now persists the whole state, including `featureModelConfigUserRevision`, which the previous code deliberately destructured out. The counter now survives a reload that the data it counts (`providerChannels`, `providerKeys`) does not, producing the `userRevision > syncedRevision` + empty-channels state above.

On versioning: leaving `version: 18` is *not* a crash risk — pre-existing blobs simply lack both keys and zustand's shallow merge falls back to `0/0`, which compares equal and short-circuits. The hazard here is semantic, not schema.

**6. `settings-dialog.tsx:4523` — medium — N1 residual**

Both new refs are per-mount, so a cleared ComfyUI baseUrl is refilled after any remount (mode switch, tab switch, dialog reopen). The user's intent to leave the field empty isn't recorded anywhere durable.

**7. `settings-dialog.tsx:4528` — low — new consequence of the ran-once flag**

If the backend re-hydration effect (`:2480-2493`) later resets `baseUrl` to the saved empty value, the self-heal cannot re-fire because `didSelfHealComfyBaseUrl.current` is already `true`, so the channel silently ends up with no baseUrl. Reachable when `savedProviderChannels` content changes (e.g. after another client saves) while the healed value was never saved locally.

**8. `settings-dialog.tsx:2099` — low**

`markAdvancedSettingsProfileSynced()` inside `handleApply` runs after several `await`ed mutations. In custom mode the advanced `FeatureModelsBlock` renders alongside the Quick panel and its inputs are not disabled during apply, so edits made while the multi-step apply is in flight bump `userRevision` and are then marked synced without ever being written into the profile JSON.

NewYuee:
已处理 Follow-up review 2 的 8 项意见，更新提交：`6116dc36`。

1. **延迟消费同步游标**：Quick profile 只有在模型网关配置加载完成、`FeatureModelsBlock` 已消费当前 `providerChannels` 后才允许恢复或同步。
2. **阻止空 ComfyUI 覆盖后端**：水合检查同时确认后端渠道、媒体模型、Embedding 和当前模型映射引用的 provider 均已存在；半水合状态不会生成或保存空渠道。
3. **记录准确的消费版本**：同步 effect 捕获 `consumedRevision` 并显式传入 `markFeatureModelConfigProfileSynced(consumedRevision)`；同步期间出现的较新编辑继续保持 pending。
4. **N4 恢复行为安全化**：保留“已有后端配置时恢复为 Custom”的产品行为，但恢复分支同样受完整水合门禁保护，不再基于启动时的半份 store 生成 applied JSON。
5. **消除持久化错配**：`userRevision`、`syncedRevision` 和渠道水合 key 均为瞬态；只持久化 `profileSyncPending`，用于刷新后在后端真源水合完成时重新同步。
6. **ComfyUI 清空意图可持久化**：用户主动清空 Base URL 后记录 durable dismiss 状态，切换模式或重开设置不会再次自动填回。载入模板或清除 ComfyUI 配置会重置该状态。
7. **恢复后续自愈能力**：移除 per-mount ran-once ref；未被用户明确关闭自动填充时，后端后续再次返回空 Base URL 仍可重新自愈。
8. **避免 Apply 吞掉并发编辑**：Apply 期间禁用模式切换和 Advanced Settings 编辑控件，并移除多段 await 结束后的无条件 mark synced。

另外自审时清理了状态文件中的无关格式化噪音，并增加持久化边界测试，确认 pending 意图保留，但修订计数、渠道水合 key、provider channel/key 草稿不会写入 localStorage。

验证结果：
- `pnpm test src/__tests__/stores/settings-store-feature-model-config.test.ts src/__tests__/lib/queries/model-gateway.test.tsx src/__tests__/i18n/locales-json.test.ts`：16/16 通过
- `pnpm test src/__tests__/stores/settings-store-feature-model-config.test.ts`：6/6 通过
- `pnpm build`：通过
- `pnpm exec tsc -b --pretty false`：通过
- `.venv/bin/pytest tests/test_p0b_compliance_generator.py::test_license_inventory_covers_current_git_index`：通过
- `git diff --check`：通过

lywaterman (commented):
## Follow-up review 3 — `6116dc36`

Reviewed `a9cadc05..6116dc36` with the surrounding code in `settings-dialog.tsx` and `settingsStore.ts`, plus the backend invariants in `newapi_provisioner.py` / `model_gateway.py`.

> **Blocking:** N1 and N2 below. Both can leave the hydration gate closed with `featureModelConfigProfileSyncPending` stuck at `true`, which silently disables the entire sync mechanism this PR exists to provide — with no timeout, no fallback, and nothing visible to the user.

### Verdicts on the second follow-up's 8 findings

| # | Finding | Verdict |
|---|---------|---------|
| F1 | Persisted cursor consumed against an empty store | **Partially fixed** — the gating is correct, but the gate itself can latch closed (N1, N2) |
| F2 | Empty-baseUrl comfyui channel overwrites the live backend | **Partially fixed** — no longer synthesized for unhydrated providers, but a legitimately empty comfyui baseUrl still corrupts the profile (N3) |
| F3 | `markFeatureModelConfigProfileSynced` recorded the latest, not the consumed, revision | **Fixed** — `consumedRevision` is captured at effect entry and passed explicitly; `pending` is recomputed as `userRevision > consumedRevision` |
| F4 | `!appliedProfileJson` recovery branch on a half-hydrated store | **Fixed** — the recovery branch is behind the same gate at `:1864` |
| F5 | `partialize` persisted `featureModelConfigUserRevision` | **Fixed** — all three revision/hydration keys are stripped; only `profileSyncPending` and `comfyUIBaseUrlAutoFillDismissed` persist. The persist boundary itself is sound (revisions reset to 0/0, `pending=true` still forces one sync pass which then clears it). The hazard is that the pending flag has no resolution path while the gate is closed — see N1/N2/N3 |
| F6 | Per-mount refs made the ComfyUI baseUrl unclearable | **Fixed** for the in-session case; residual cross-device gap (N6) |
| F7 | Ran-once flag prevented later self-heal | **Fixed** — the ref is gone and the effect re-arms on `channel?.settings` / dismissal changes |
| F8 | `handleApply` marked synced after awaits while advanced inputs stayed editable | **Mostly fixed** — the unconditional mark is removed and the tabs plus the `fieldset` are disabled; one escape hatch remains (N5) |

### New findings

**N1. `settings-dialog.tsx:1774` — high — the hydration gate can latch closed permanently**

`isQuickProfileAdvancedSettingsHydrated` requires every provider named by `config.provisioner.mediaModels`, `config.provisioner.embeddingModel`, and the locally persisted `featureModels` / `mediaModels` to have an entry in `settings.providerChannels` — but that map is only ever populated from `config.provisioner.providerChannels` (`:2531`).

The backend enforces no such invariant:

- `providerChannels`, `mediaModels` and `embeddingModel` are three independent `runtime_settings` rows (`src/novelvideo/newapi_provisioner.py:1541-1543`).
- `POST /provider-channels` omitting a provider skips the media-mapping reconciliation entirely (`src/novelvideo/api/routes/model_gateway.py:800-850`).
- `embeddingModel.provider` is never cleared when its channel goes away.
- `providerChannels: []` alongside a non-empty `mediaModels` is an asserted-reachable state (`tests/test_model_gateway_settings.py:2812`, after a failed channel sync).

With a single orphaned media model, `advancedSettingsHydrated` is `false` on every render forever: the sync effect and the F4 recovery branch never run, `featureModelConfigProfileSyncPending` stays `true` in localStorage indefinitely, and there is no timeout, fallback or user-visible signal. The feature is silently dead.

**N2. `settings-dialog.tsx:4888` — high — deleting a provider-channel row closes the gate for the rest of the visit**

The row's delete button calls `removeFeatureProviderChannel(provider)`, which cascades in the local store (dropping that provider's `featureModels` / `mediaModels` / `embeddingModel`) but cannot touch the backend snapshot. `config.provisioner.providerChannels` still lists the provider, so it stays in `requiredProviders` while `settings.providerChannels[provider]` is now gone → gate `false`.

`FeatureModelsBlock`'s hydration effect will not re-add it, because its per-mount `lastSyncedProviderChannelsKey` ref already equals the unchanged `savedProviderChannelsKey` (`:2536`). The removal also bumps the revision and sets `pending = true`, so every subsequent Advanced Settings edit is recorded as pending and never reaches the quick profile — until a config refetch changes the channel key or the block remounts. Nothing tells the user.

**N3. `settings-dialog.tsx:1864` — medium — F6's fix now lets an empty ComfyUI baseUrl corrupt the applied profile permanently**

The gate only checks that `settings.providerChannels.comfyui` *exists*, not that its `baseUrl` is usable. `syncQuickProfileFromAdvancedSettings`'s `for (const channel of Object.values(settings.providerChannels))` loop unconditionally writes `target.baseUrl = channel.baseUrl` (`:1675-1682`).

So clearing the ComfyUI base URL in Advanced Settings — now that it actually sticks — produces a `syncedJson` containing `comfyui.baseUrl: ""`, which the effect writes into `customProfileJson`, `appliedProfileJson` and localStorage. `parseQuickModelProfile` then throws `ComfyUI channel.baseUrl is required` (`:1595`): `parsedProfile` becomes `null`, Apply fails with "invalid JSON", and the effect's own `parseQuickModelProfile(appliedProfileJson)` hits `catch { return; }` at `:1897` *before* marking synced — so `pending` stays true and the corrupt profile survives every reload. The sync should skip or repair channels that would not round-trip through `parseQuickModelProfile`.

**N4. `frontend/src/__tests__/stores/settings-store-feature-model-config.test.ts:113` — medium — the new persistence test doesn't assert what it claims**

`expect(persisted.featureModelConfig).toMatchObject({ providerChannels: {}, providerKeys: {} })` is vacuous: `toMatchObject` subset-matches nested objects, so an empty expected object matches *any* object. The test sets up a real comfyui channel via `updateFeatureProviderChannel` (which creates the channel — `settingsStore.ts:751`), yet would still pass if `prepareFeatureModelSettingsForPersistence` stopped stripping channels and upstream keys — i.e. it gives no protection against leaking API keys into localStorage. Use `toEqual({})` on the two sub-objects.

The other four new tests do assert their claims correctly — the F3 "older revision, newer edits stay pending" one in particular.

**N5. `settings-dialog.tsx:536` — low — the apply lock is escapable via the settings nav**

`quickProfileApplying` disables the mode tabs and the Advanced `fieldset`, but the "Storage" nav button (`:178`) and the dialog Close button are not disabled. Clicking Storage mid-apply unmounts `ModelConfigSection`; clicking back to Models remounts it with fresh `useMutation` instances whose `isPending` is `false`, so `quickProfileApplying` is `false` and Advanced Settings is editable again while the apply is still in flight. Edits made in that window bump the revision, then the in-flight apply's post-await store writes (`:2124-2161`, all `{ source: "profile" }`) silently overwrite them without bumping it, and the sync effect marks the lost revision as consumed.

Related: the `onApplyingChange` effect at `:1938-1940` has no cleanup, so it also cannot report `false` on unmount.

**N6. `settings-dialog.tsx:4833` — low — the ComfyUI dismiss flag is client-local while the data it guards is server-side**

`comfyUIBaseUrlAutoFillDismissed` lives in localStorage, but the ComfyUI baseUrl it protects lives in the backend channel. On a second browser, a different machine, or after clearing site data, `dismissed` is `false` again, so the self-heal effect (`:4599-4617`) re-fills `http://127.0.0.1:8188` into a deliberately cleared baseUrl and resurrects it on the next save.

It is also a single provider-global boolean rather than per-channel — harmless today since only `comfyui` reads it, but it will collide if another provider gains an autofill.

NewYuee:
已将 Follow-up review 3 收敛为统一的数据流实现，更新提交：`694957d4`。

- 后端 provider/media/embedding 以单个快照原子 hydrate；同步门禁只判断该快照是否已消费，不再因孤儿引用或本地删除永久关闭。
- profile 统一通过纯 reconciler 生成：忽略孤儿媒体模型，排除空/无 workflow 的 ComfyUI；当前配置无法形成合法 profile 时保留最后一份可用配置并消费 revision，不写入损坏 JSON。
- 后端拥有的渠道、媒体模型和 embedding 不再持久化到 localStorage，仅保留前端功能模型映射，避免半水合状态和跨端分叉。
- 移除客户端 ComfyUI 自动回填/单机 dismiss 状态；载入模板仍会明确填入默认地址，原模板能力不变。
- Apply 锁提升到设置弹窗父级，并覆盖整个多步请求：期间禁止切页、关闭、切模式和编辑所有模型配置，异常路径在 finally 统一释放。
- 修正持久化测试的空对象断言，并补充后端原子水合、本地删除、孤儿 media/embedding、空 ComfyUI 和无有效渠道等回归用例。

自审结论：未发现新的行为回归；官方/自定义/混合模式以及原有渠道、模型、Embedding、媒体模型、ComfyUI 模板配置能力均保留。

验证：
- `pnpm build`：通过
- `pnpm exec vitest run src/__tests__/stores/settings-store-feature-model-config.test.ts src/__tests__/lib/queries/model-gateway.test.tsx src/__tests__/i18n/locales-json.test.ts`：22/22 通过
- `.venv/bin/pytest tests/test_p0b_compliance_generator.py::test_license_inventory_covers_current_git_index`：通过
- `git diff --check`：通过
- 完整 `pnpm test`：2283/2289 通过；其余 6 项仍为当前 Node 环境的 localStorage 与 FormData/MSW 兼容失败，失败文件不在本次修改范围内。

frontend/src/components/settings/settings-dialog.tsx
```diff
@@ -2399,100 +2592,84 @@ function FeatureModelsBlock({
...
+  );
+  const lastHydratedBackendSnapshotKey = useRef("");
   useEffect(() => {
-    if (!defaultComfyWorkflows) return;
-    const current = providerChannels.comfyui;
-    if (!current || current.baseUrl.trim()) return;
-    const workflows = readComfyUIWorkflows(current.settings);
-    const hasLoadedTemplate = Object.keys(defaultComfyWorkflows.workflows).some(
-      (model) => workflows[model],
+    if (!backendSnapshotLoaded) return;
```
lywaterman, line 2602: **medium — hydration now discards unsaved provider-channel edits, and is triggered by saves that have nothing to do with channels.**

Two changes compound here:

1. `backendSnapshotKey` (line 2596) now folds `savedMediaModels` and `savedEmbeddingModel` into the key, so the effect re-fires on media/embedding saves too.
2. `hydrateFeatureModelConfigFromBackend` replaces `featureModelConfig.providerChannels` **wholesale** with the backend list.

Concrete scenario: in Advanced Settings the user edits the ComfyUI channel's Base URL / workflow JSON (store-only, not yet saved), then clicks **Save** in the Media Models section. `useSaveMediaModels` invalidates `queryKeys.modelGateway()`, `config.provisioner.mediaModels` changes, the snapshot key changes, and the effect re-hydrates — silently discarding the unsaved channel edits, and dropping any channel the user added locally but has not persisted.

The code this replaced (`lastSyncedProviderChannelsKey`) keyed only on `savedProviderChannels` and only added/patched channels — it never removed them and never fired on a media/embedding save. The wholesale replacement is the right shape for correctness, but it needs to either key on `savedProviderChannels` alone, or preserve locally-dirty channels across a hydrate.

frontend/src/components/settings/settings-dialog.tsx
```diff
@@ -4242,14 +4438,22 @@ function ProviderChannelsBlock({
...
-            (!allowedProviderSet || allowedProviderSet.has(provider)) &&
-            !excludedProviderSet.has(provider),
-        ).map((provider) => ({
-          provider,
-          name: featureProviderLabel(provider),
-        }));
+      : Array.from(
+          new Set([
+            ...FEATURE_MODEL_PROVIDERS,
+            ...Object.keys(CUSTOM_PROVIDER_CHANNEL_TYPES),
```
lywaterman, line 4444: **medium — the `CUSTOM_PROVIDER_CHANNEL_TYPES` union is only applied in the fallback branch, which makes ComfyUI unaddable and the relocated template button unreachable.**

The union that injects `comfyui` / `fal_ai` into `providerOptions` sits in the `channelTypes.length === 0` branch. The `channelTypes.length > 0` branch (line 4435) still filters the API list only. This PR also adds `customProviderChannelType()` as a `type` fallback in three save paths (lines 1717, 2695, 4509), which implies `comfyui` and `fal_ai` are **not** returned by `/api/channel/types`.

Chain in Hybrid mode:

- `FeatureModelsBlock` is rendered with `comfyOnly` (line 559) → `allowedProviders={["comfyui"]}` (line 2778)
- `channelTypesEnabled` is true once Local NewAPI is configured and the provisioner DB is available (line 543)
- once the query resolves non-empty, `providerOptions` → `[]`, `availableProviders` → `[]`, and `handleAdd` early-returns — the **Add Channel** control does nothing

And because this PR removed the top-level template button (which called `addFeatureProviderChannel("comfyui")` itself) and moved it inside `ProviderChannelRow` / `ComfyUIWorkflowsEditor` (line 4964, reachable only from an existing channel row), a user with no ComfyUI channel now has no path to create one or reach **Load MiniMax H3 Template**. That directly contradicts the new docs ("Click Add Channel and select ComfyUI").

Fix: apply the same union in both branches, e.g. merge `Object.keys(CUSTOM_PROVIDER_CHANNEL_TYPES)` into the API-derived list before filtering.

Please confirm whether `/api/channel/types` returns type 63 — if it does, this is a non-issue and only the fallback asymmetry remains.

frontend/src/components/settings/settings-dialog.tsx
```diff
@@ -1797,32 +1958,51 @@ function QuickLocalNewApiSetup({
...
+    if (!syncedProfile) {
+      markAdvancedSettingsProfileSynced(consumedRevision);
+      return;
+    }
+    const syncedJson = JSON.stringify(syncedProfile, null, 2);
+    if (syncedJson === appliedProfileJson) {
+      markAdvancedSettingsProfileSynced(consumedRevision);
+      return;
+    }
+    setSelectedProfileKind("custom");
```
lywaterman, line 1989: **low — this unconditionally overwrites a hand-written "My Config" and force-switches away from "Recommended".**

The effect does `setSelectedProfileKind("custom")` plus `saveStoredQuickProfiles({ selected: "custom", customProfileJson: syncedJson, ... })` with no guard on what the user currently has.

Scenario: the user has a hand-written custom profile saved in `dramaclaw-ce-quick-model-profiles` but is currently on **Recommended**. Any Advanced Settings edit — or any backend snapshot change that flips `featureModelConfigProfileSyncPending` — overwrites `customProfileJson` in localStorage with the generated JSON, permanently losing the hand-written profile, and silently moves the selector to **My Config**.

The previous code at least preserved `selected: selectedProfileKind`. Suggest gating the overwrite: only rewrite `customProfileJson` when the user is already on `"custom"`, or when the stored value equals the last generated one.

frontend/src/components/settings/settings-dialog.tsx
```diff
@@ -1370,20 +1417,29 @@ const HYBRID_COMFYUI_WORKFLOWS = {
...
+  fal_ai: 61,
+  comfyui: 63,
+};
+
+function customProviderChannelType(provider: string): number | undefined {
+  return CUSTOM_PROVIDER_CHANNEL_TYPES[provider];
+}
+
 const RECOMMENDED_MEDIA_MODELS: Readonly<Record<string, QuickProfileModel>> = {
   "LingShan-G2": {
```
lywaterman, line 1430: **low — inconsistent fal.ai model identifier form in `RECOMMENDED_MEDIA_MODELS`.**

`LingShan-G2 → "gpt-image-2"` and `LingShan-NB-2 → "nano-banana-2"` are mapped to the new `fal_ai` channel as bare ids, while the audio entries in the same table use fully-qualified fal endpoint ids (`fal-ai/index-tts-2/text-to-speech`, `fal-ai/elevenlabs/music`).

If fal requires the namespaced form, applying the recommended profile writes a model mapping that resolves to model-not-found at generation time for exactly those two image models — and it fails at generation, not at Apply, so it would not show up in the validation you ran. Worth confirming both forms are valid for the type-61 channel.

frontend/src/stores/settingsStore.ts
```diff
@@ -543,48 +670,59 @@ export const useSettingsStore = create<SettingsState>()(
...
             embeddingModel: model
               ? {
                   provider: normalizeFeatureModelProvider(model.provider),
-                  upstreamModel: model.upstreamModel.trim(),
+                  upstreamModel:
+                    typeof model.upstreamModel === 'string'
+                      ? model.upstreamModel.trim()
+                      : '',
                   dimension: Math.max(1, Math.round(Number(model.dimension) || 0)),
+                  ...(model.batchSize != null
```
lywaterman, line 713: **low — `setEmbeddingModel` coerces an invalid `batchSize` to `1` instead of dropping it.**

```ts
...(model.batchSize != null
  ? { batchSize: Math.max(1, Math.round(Number(model.batchSize) || 0)) }
  : {})
```

`Math.max(1, ...)` turns `0` (and `NaN`) into `1`, which disagrees with `normalizeFeatureModelSettings` — that one omits non-finite / non-positive values.

Concrete path: `EmbeddingModelBlock.handleSave` calls `setEmbeddingModel({ ..., batchSize: res.data.embeddingModel.batchSize })`. If the backend returns `0` to mean "unset", the store records `batchSize: 1`, which then flows into the generated quick profile and gets written back to the backend as a real batch size of 1 on the next Apply — an unset value silently becomes a configured one.

Matching `normalizeFeatureModelSettings` (omit unless finite and > 0) would keep the two paths consistent.

lywaterman (commented):
## Follow-up review 4 — `694957d4`

The architectural convergence is the right call. Reviewed the full diff against `main` this round rather than only the delta, since the rework touches most of the previous surface.

**What I verified as genuinely fixed:**

- The hydration gate can no longer latch closed. `featureModelConfigProfileSyncPending` and `featureModelConfigBackendSnapshotKey` are both stripped in `partialize`, and `comfyUIBaseUrlAutoFillDismissed` is gone entirely — so there is no persisted state that can survive a reload and permanently disable sync. This resolves the blocking N1/N2 from the last round.
- Backend-owned state (`providerChannels`, `providerKeys`, `mediaModels`, `embeddingModel`) is no longer persisted at all, which removes the half-hydrated-store class of bug rather than patching instances of it.
- Revision bookkeeping (`nextFeatureModelConfigMutationMeta` / `markFeatureModelConfigProfileSynced`) converges and does not loop — after `markSynced` the selectors return equal primitives, so no re-render is triggered.
- `parseQuickModelProfile(JSON.stringify(next))` in `syncQuickProfileFromAdvancedSettings` cannot throw for well-formed store state; the provider regex, uniqueness, positive dimension/batchSize and ComfyUI baseUrl/model/workflow cases are all pre-filtered.
- `onApplyingChange(true)` is placed after every early `return` in `handleApply`, and the `finally` plus the unmount cleanup make the "dialog cannot be closed" state non-deadlockable.

### New findings

| # | Location | Severity |
|---|----------|----------|
| 1 | `settings-dialog.tsx:2602` | medium — hydration wipes unsaved channel edits, and fires on unrelated saves |
| 2 | `settings-dialog.tsx:4433` | medium — ComfyUI cannot be added once `/api/channel/types` loads, making the relocated template button unreachable |
| 3 | `settings-dialog.tsx:1989` | low — profile sync overwrites a hand-written "My Config" and force-switches off "Recommended" |
| 4 | `settings-dialog.tsx:1430` | low — inconsistent fal.ai model id form in the recommended profile |
| 5 | `settingsStore.ts:713` | low — `setEmbeddingModel` coerces an invalid `batchSize` to `1` |

Finding 2 is the one I would not ship without resolving — please confirm whether `/api/channel/types` actually returns `comfyui` (type 63); the existence of `customProviderChannelType()` as a fallback in three save paths suggests it does not.

Details inline. Minor non-finding: the duplicate-id suffix loop in `syncQuickProfileFromAdvancedSettings` is dead code (`providerChannels` is keyed by provider), but harmless.

frontend/src/components/settings/settings-dialog.tsx
NewYuee: 感谢指出。这个场景要求用户在渠道 Base URL/Workflow 仍处于未保存草稿时，转而保存 Media Models 或 Embedding，才会被后端快照刷新覆盖。当前配置流程约定各区块独立保存，正常操作是先保存渠道配置再编辑其他区块；按当前使用方式不会影响现有功能。本 PR 不再为跨区块未保存草稿增加冲突合并状态，接受该边界，本项不处理。

frontend/src/components/settings/settings-dialog.tsx
NewYuee: 这是当前明确的产品行为：高级配置发生变化后自动切换为 Custom，并根据数据库中的渠道和模型重新生成 JSON。保留手写但未应用的 Custom 草稿会与“高级配置是真源、自动重建 JSON”的既定语义冲突。当前行为符合本功能需求，本项不处理。

frontend/src/components/settings/settings-dialog.tsx
NewYuee: 和  是当前 RelayClaw/DramaClaw 已使用的模型映射与兼容别名，且该映射不是本轮同步收敛改动引入。当前没有实际请求失败或上游要求必须改为 namespaced ID 的证据；贸然替换可能破坏现有渠道映射。本项不处理，后续如有真实渠道错误再单独调整。

frontend/src/stores/settingsStore.ts
NewYuee: 当前前端输入与后端保存校验都只允许正整数，后端  非正数会直接拒绝，不会在正常配置链路中返回  或 。因此该防御性差异不影响当前功能；本 PR 不扩大修改范围，本项不处理。

frontend/src/components/settings/settings-dialog.tsx
NewYuee: 已核对当前实际依赖的 RelayClaw CE，而不是按通用 NewAPI 假设判断。RelayClaw CE 的 GetRegisteredChannelTypes 明确注册 ChannelTypeComfyUI=63，relay/channel_types_test.go 和 controller/channel_types_test.go 也分别断言 provider=comfyui 会出现在 /api/channel/types 返回中。因此接口正常加载时 channelTypes 已包含 ComfyUI，Hybrid 的 Add Channel 和模板入口均可达；该场景在当前支持的网关中不成立，本项不处理。

lywaterman (approved):

PR #309, opened 2026-08-18
fix: preserve scoped project state paths

Handanhhhy:
## Summary

- pass the registry-resolved `ProjectContext.state_dir` into five store-backed task runners
- prevent scoped projects from falling back to a path derived only from `owner/project`
- cover all affected runtime construction sites with a registry-scoped regression test

## Root cause

The affected runners passed only `owner/project` and `output_dir` to `CogneeStore`. A project state
directory may contain registry-owned scope segments that cannot be reconstructed from that label,
so the fallback could open and initialize a different, empty database.

Affected paths:

- character and identity images
- script writer
- scene reference assets
- prop reference assets
- projection-free sketch scene lookup

`beat_video_prompt` is not changed because staging already uses the context-aware SQLite factory.

## Verification

- RED on staging: `tests/test_task_runner_state_dir.py` failed in all five parameterized cases with
  missing `state_dir`
- GREEN after fix: all five parameterized cases pass
- focused runner and compliance verification: `31 passed`
- Ruff: `All checks passed!`
- complete test coverage in clean processes: the isolated order-sensitive project-summary test
  passed, and the remaining suite reported `3199 passed, 16 skipped, 2 deselected`

The repository's single-process suite still exposes an unrelated pre-existing order dependency:
`test_project_summary_counts_from_registered_state_dir` returns null counts after earlier modules,
but passes in a clean process. This PR does not modify that route or its dependencies.

## Compatibility

- no schema migration
- no API or configuration change
- default project paths retain the same resolved state directory
- no existing fallback directories are deleted by this PR


src/novelvideo/task_backend/runners/script.py
```diff
@@ -117,7 +117,11 @@ def update_progress(progress: float, task: str) -> None:
...
 
-    store = CogneeStore(ctx.owner_project_label, output_dir=output_dir)
+    store = CogneeStore(
+        ctx.owner_project_label,
+        output_dir=output_dir,
+        state_dir=str(ctx.state_dir),
+    )
     await store.initialize()
     try:
         project_config = load_project_config(ctx.owner_username, ctx.project_name)
```
lywaterman: **medium — same non-scoped fallback this PR is fixing, two lines below the fix.**

`load_project_config(ctx.owner_username, ctx.project_name)` resolves through `get_project_config_path()` → `STATE_DIR/<owner>/<project>/project_config.json` (`src/novelvideo/project_config.py:48-56`) — exactly the owner/project-derived path the store no longer uses.

The write side is already state_dir-authoritative: `src/novelvideo/api/routes/projects.py:579,772` only ever call `save_project_config_in_state_dir(ctx.state_dir, ...)`.

Scenario: an org project at `STATE_DIR/_orgs/org_1/alice/demo` has `genre` and `story_setting` configured via the API. `script_writer` now opens the correct DB, but `load_project_config` reads a path where no `project_config.json` exists, gets `{}`, and `merged_config.get("genre", "")` / `get("story_setting", "")` / `get("spine_template", "drama")` all silently fall back to defaults — the workflow generates a script that ignores the project's configuration, with no error.

`runners/episode_assets.py:123` already shows the correct call:

```python
project_config = load_project_config_file_from_state_dir(ctx.state_dir)
```

src/novelvideo/task_backend/runners/character_image.py
```diff
@@ -129,7 +129,11 @@ def update(progress: float, current_task: str) -> None:
         )
 
     update(0.10, "加载角色数据...")
-    store = CogneeStore(ctx.owner_project_label, output_dir=str(output_dir))
+    store = CogneeStore(
+        ctx.owner_project_label,
+        output_dir=str(output_dir),
+        state_dir=str(ctx.state_dir),
+    )
```
lywaterman, line 136: **medium — same defect at line 143, seven lines below this fix.**

```python
project_config = load_project_config_file(ctx.owner_username, ctx.project_name)
ethnicity = project_config.get("ethnicity", "Chinese")
```

For an org-scoped project the config file lives at `ctx.state_dir/project_config.json`, so this owner/project-derived load returns `{}` and every portrait/identity image is generated with `ethnicity="Chinese"` regardless of what the project was actually configured with. Silent, and visible only in the output images.

Should be `load_project_config_file_from_state_dir(ctx.state_dir)`, matching `runners/episode_assets.py:123`.

docs/superpowers/specs/2026-08-18-org-runner-state-dir-design.md
```diff
@@ -0,0 +1,85 @@
...
+output directory or changing the store fallback would make tenant identity implicit and could mask
+future wiring errors.
+
+## Data Flow After the Fix
+
+1. The task consumer resolves the project record into `ProjectContext`.
+2. `ctx.state_dir` retains the organization-qualified registry path.
+3. The runner passes that exact path to `CogneeStore`.
+4. `CogneeStore` and its `SQLiteStore` use `<ctx.state_dir>/data.db`.
+5. No personal-style fallback directory is created during organization task execution.
```
lywaterman: **low — this claim doesn't match what the code does.**

In `src/novelvideo/cognee/store.py`, the `default_state_dir` if/elif chain runs *unconditionally*, before the `if state_dir:` override is consulted. Because `project_name` is `owner/project`, it still takes the `elif "/" in project_name` branch and calls `ProjectPaths(owner, project).bootstrap_from_legacy_output()`, which `mkdir`s `STATE_DIR/<owner>/<project>` and touches the migration marker even when nothing is migrated.

So every org task still creates an empty personal-style state directory — and if a same-named legacy personal payload exists under `OUTPUT_DIR/<owner>/<project>`, it gets copied there.

Not introduced by this PR, but the doc asserts the opposite. Either short-circuit the fallback computation when `state_dir` is supplied, or soften the claim to match current behavior.

lywaterman (commented):
## Review

**The core change is correct.** Passing `state_dir=str(ctx.state_dir)` into `CogneeStore` is the right fix: `CogneeStore.__init__` uses the explicit `state_dir` for both `data.db` and the Cognee/Kuzu roots (`cognee_project_context(state_dir)` in `src/novelvideo/cognee/ladybug_access.py`), so SQLite and the graph stay in the same tenant scope.

I checked every other `CogneeStore(...)` construction under `task_backend/runners/` (`ingest`, `graph_build`, `episode_assets`, `identity`, `video`) — they already pass `state_dir`, so this PR closes the last five gaps and none are missed. The new parameterized test genuinely reaches all five sites (the runners import `CogneeStore` lazily inside the function, so `monkeypatch.setattr(novelvideo.cognee, "CogneeStore", ...)` takes effect), and it asserts on the right value. `ProjectContext.state_dir` is a required non-optional `Path`, so `str(...)` is safe at every site.

Three findings below — two are the *same class of bug* as the one being fixed, sitting a few lines away from the fixed call sites.

| # | File | Severity |
|---|------|----------|
| 1 | `runners/script.py:127` | medium |
| 2 | `runners/character_image.py:143` | medium |
| 3 | design doc, data-flow claim 5 | low |

The runner edits themselves are behavior-preserving for personal projects (`ctx.state_dir` equals the previously derived path), and the updated assertion in `tests/test_sketch_runner_reads_projection.py` matches the new kwargs.

Handanhhhy:
@lywaterman The three review points are addressed in cc3d81c1:

- script and character-image runners now load project config from `ctx.state_dir`
- explicit `state_dir` short-circuits fallback path derivation in both `CogneeStore` and `SQLiteStore`
- regression coverage was added with RED→GREEN verification

Local verification: 90 focused tests passed; the full suite passed in split clean processes (3201 tests plus the isolated order-dependent test); Ruff and diff checks passed. The branch was also squashed to remove scope-specific deployment details from the public history. Please re-review when convenient.

src/novelvideo/sqlite_store.py
```diff
@@ -339,7 +339,9 @@ def __init__(
             self.project_dir = ensure_project_dirs(project_name)["base"]
 
         parts = project_name.split("/", 1)
-        if len(parts) == 2:
+        if state_dir:
```
lywaterman, line 342: **high — this disables legacy-layout migration for every caller that passes `state_dir`.**

The new `if state_dir:` short-circuits the `elif len(parts) == 2:` branch, which was the only call site of `ProjectPaths(...).bootstrap_from_legacy_output()` here. Note that `resolved_state_dir` is *already* recomputed as `Path(state_dir)` ten lines below — so the old branch existed purely for that side effect, and the assignment this change adds is dead.

Verified on both revisions (legacy `output/<user>/<proj>/data.db`, no `.migrated` marker, `state_dir=STATE_DIR/<user>/<proj>`):

```
base 4a0a1a8a → LEGACY DATA MIGRATED: True
head cc3d81c1 → LEGACY DATA MIGRATED: False
```

Scenario: an installation upgrading from the old single-directory layout. Any un-migrated project whose first post-upgrade access goes through `api/deps.py:make_sqlite_store` / `make_sqlite_store_for_context` — both of which already passed `state_dir` before this PR — now gets a freshly created, empty `data.db` in the state dir instead of the migrated one.

Suggestion: gate on whether the path is genuinely scope-owned rather than on truthiness, e.g. still run the derived branch when `Path(state_dir).resolve() == ProjectPaths(parts[0], parts[1]).state_dir.resolve()`.

src/novelvideo/cognee/store.py
```diff
@@ -144,7 +144,9 @@ def __init__(
 
             self.project_dir = ensure_project_dirs(project_name)["base"]
 
-        if sqlite_state_dir:
+        if state_dir:
```
lywaterman, line 147: **high — identical regression to `sqlite_store.py:342`.**

`if state_dir:` skips the `elif "/" in project_name:` branch and with it `bootstrap_from_legacy_output()`; `resolved_state_dir` is recomputed from `state_dir` at line 170 regardless, so the added assignment is dead and the removed side effect is the entire behavior change.

Verified with a constructor-only repro: base `True`, head `False`.

This affects `make_cognee_store`, `make_cognee_store_for_context`, and the already-converted runners (`ingest`, `graph_build`, `episode_assets`, `identity`, `video`) — none of which this PR touched.

Same suggested fix as `sqlite_store.py`: skip the derived branch only when the supplied path is not `ProjectPaths(owner, project).state_dir`.

src/novelvideo/task_backend/runners/script.py
```diff
@@ -117,10 +117,14 @@ def update_progress(progress: float, task: str) -> None:
...
-    store = CogneeStore(ctx.owner_project_label, output_dir=output_dir)
+    store = CogneeStore(
+        ctx.owner_project_label,
+        output_dir=output_dir,
+        state_dir=str(ctx.state_dir),
+    )
     await store.initialize()
     try:
-        project_config = load_project_config(ctx.owner_username, ctx.project_name)
+        project_config = load_project_config_file_from_state_dir(ctx.state_dir)
```
lywaterman: **medium — two issues with this swap.**

**(a) Legacy projects now read empty.** `load_project_config(username, project)` went through `get_project_config_path()`, which calls `bootstrap_from_legacy_output()` before returning the path. The new call has no such bootstrap. Reproduced: for an un-migrated project with `output/<u>/<p>/project_config.json` = `{"ethnicity":"Korean","spine_template":"narrated"}`, `load_project_config_file_from_state_dir(state_dir)` returns `{}` while `load_project_config_file(u, p)` returns the real dict. Result: `create_script_writing_workflow` runs with `genre=""`, `story_setting=""`, `spine_template="drama"` — a project configured as `narrated` silently generates the wrong spine.

**(b) Wrong tier of loader.** `load_project_config` applied `_effective_project_config` (defaults merge, `video_backend` migration, `visual_style` validation) plus the `project_uuid` backfill; `load_project_config_file_from_state_dir` is a raw file read. The codebase already has the exact state-dir equivalent:

```python
project_config = load_project_config_from_state_dir(
    ctx.state_dir,
    username=ctx.owner_username,
    project=ctx.project_name,
)
```

For the three keys read here the values happen to coincide today (`spine_template` default is `"drama"` on both sides), so this is drift risk rather than a live bug — but it's a free correctness win to use the matching loader.

src/novelvideo/task_backend/runners/character_image.py
```diff
@@ -129,14 +129,18 @@ def update(progress: float, current_task: str) -> None:
...
     await store.initialize()
     await store.load_graph_state()
     try:
         character = await store.get_character_from_graph(character_name)
         if character is None:
             raise RuntimeError(f"找不到角色: {character_name}")
-        project_config = load_project_config_file(ctx.owner_username, ctx.project_name)
+        project_config = load_project_config_file_from_state_dir(ctx.state_dir)
```
lywaterman, line 143: **medium — same legacy-config loss as `script.py:127`.**

`load_project_config_file(username, project)` bootstrapped legacy layout via `get_project_config_path()`; the state-dir variant does not. For an un-migrated legacy project the config reads `{}` and `ethnicity` falls through to the hardcoded `"Chinese"`, so portrait/identity generation silently ignores a configured ethnicity.

Correct for scoped projects, regressive for legacy ones. The conditional-bootstrap fix suggested on `sqlite_store.py` / `cognee/store.py` resolves this too; otherwise this line needs its own migration trigger.

src/novelvideo/task_backend/runners/scene_reference.py
```diff
@@ -78,7 +78,11 @@ def update(progress: float, current_task: str) -> None:
...
     update(0.10, "加载场景数据...")
-    store = CogneeStore(ctx.owner_project_label, output_dir=str(output_dir))
+    store = CogneeStore(
+        ctx.owner_project_label,
+        output_dir=str(output_dir),
+        state_dir=str(ctx.state_dir),
+    )
     await store.initialize()
     try:
         scene = await store.sqlite_store.get_scene(scene_name)
```
lywaterman, line 88: **low — scoped-path gap left unfixed at line 97, in a function this PR touched.**

```python
preset = get_style_preset(
    style_id,
    username=ctx.owner_username,
    project=ctx.project_name,
    ...
)
```

This reaches `StyleService._load_project_custom_style_map` → `load_project_config_file(username, project)`, i.e. `STATE_DIR/<owner>/<project>/project_config.json`. For a registry-scoped project whose real config lives under the scope-qualified path, custom styles are not found and the runner silently falls back to the system preset — the same class of bug this PR fixes for the store, nine lines below the changed hunk.

Out of scope if you'd rather do it separately, but worth a follow-up issue so it isn't lost.

lywaterman (commented):
## Re-review of `cc3d81c1`

Thanks — all three points from the previous round are addressed, and the runner wiring is now complete: I re-checked every `CogneeStore(` / `SQLiteStore(` construction site in `src/` and none is missing `state_dir`. `tests/test_task_runner_state_dir.py`, `test_sketch_runner_reads_projection.py` and `test_task_script_runner_store_lifecycle.py` pass locally (18 passed).

**However, the two-line short-circuit in the store constructors has a side effect that reaches far beyond the five runners, and I believe it is a release blocker.**

| # | File | Severity |
|---|------|----------|
| 1 | `sqlite_store.py:342` | **high** — legacy migration disabled for every caller passing `state_dir` |
| 2 | `cognee/store.py:147` | **high** — same regression in `CogneeStore` |
| 3 | `runners/script.py:127` | medium — legacy project config reads empty; also drops the defaults layer |
| 4 | `runners/character_image.py:143` | medium — same, `ethnicity` falls back to hardcoded `"Chinese"` |
| 5 | `runners/scene_reference.py:97` | low — scoped-path gap left unfixed in a touched function |

### Why 1 and 2 are high

The pre-existing derived branch existed **purely for its side effect**: `resolved_state_dir` was already `Path(state_dir)` ten lines below, so the `elif` branch's only remaining job was calling `ProjectPaths(...).bootstrap_from_legacy_output()`. The new `if state_dir:` branch removes that call and adds an assignment that is otherwise dead.

I verified this empirically on both revisions — legacy payload at `output/<user>/<proj>/data.db`, no `.migrated` marker, `state_dir=STATE_DIR/<user>/<proj>`:

```
base 4a0a1a8a   → LEGACY DATA MIGRATED: True
head cc3d81c1   → LEGACY DATA MIGRATED: False
```

This is exactly the "opens a different, empty database" failure the PR set out to prevent, reintroduced for legacy projects — and it hits paths the PR never touched (`api/deps.py:make_sqlite_store`, `make_sqlite_store_for_context`, `make_cognee_store`, `make_cognee_store_for_context`, and the already-converted runners `ingest` / `graph_build` / `episode_assets` / `identity` / `video`), all of which were already passing `state_dir` before this PR.

### Suggested shape

Rather than skipping the derived branch whenever `state_dir` is truthy, skip it only when the supplied path is genuinely scope-owned — i.e. keep running the bootstrap when `Path(state_dir)` resolves to `ProjectPaths(owner, project).state_dir`. That preserves the fix for scoped projects and keeps legacy migration intact for personal ones, and it also resolves findings 3 and 4 as a side effect.

Findings 3-5 are inline.

Handanhhhy:
@lywaterman The re-review findings are addressed in c65f67d2:

- `CogneeStore` and `SQLiteStore` preserve legacy bootstrap when an explicit path equals the personal default state path, while scope-owned paths still skip the derived fallback
- `script.py` now uses `load_project_config_from_state_dir(...)` to preserve effective-config semantics
- character-image legacy config is restored by the conditional bootstrap
- scene custom-style lookup now carries the authoritative `state_dir` through `get_style_preset` and `StyleService`

TDD coverage includes both sides of the path invariant, effective config UUID backfill, and scoped custom-style resolution. Verification: 111 focused tests passed; full split suite passed (3205 tests plus the isolated order-dependent test); Ruff and diff checks passed. Please re-review when convenient.

src/novelvideo/task_backend/runners/sketch.py
```diff
@@ -166,7 +166,11 @@ async def _ensure_scene_refs_for_beats(
     projected_scenes = None
     store = None
     if projection is None:
-        store = CogneeStore(ctx.owner_project_label, output_dir=output_dir)
+        store = CogneeStore(
+            ctx.owner_project_label,
+            output_dir=output_dir,
+            state_dir=str(ctx.state_dir),
+        )
```
lywaterman: **medium — this store is never closed, and the PR makes the leak matter.**

`_ensure_scene_refs_for_beats` constructs the store here, calls `initialize()` and `load_graph_state()`, then returns through several paths without ever calling `await store.close()`. Every sibling runner wraps it properly — `prop_reference.py:93`, `scene_reference.py`, `character_image.py:194` all use `try/finally: await store.close()`.

Each sketch task therefore leaks an open aiosqlite connection to `data.db`, and `SQLiteStore` documents itself as a one-shot lifecycle object.

The leak pre-dates this PR, but the diff changes its impact: the store now resolves to the **real scoped project DB** rather than a possibly-empty fallback, so the leaked connection now holds the live project database open for the worker's lifetime instead of an orphaned one. Given the PR already touches this exact construction, wrapping it in `try/finally` here is cheap.

```python
store = CogneeStore(
    ctx.owner_project_label,
    output_dir=output_dir,
    state_dir=str(ctx.state_dir),
)
await store.initialize()
await store.load_graph_state()
try:
    ...
finally:
    await store.close()
```

src/novelvideo/task_backend/runners/prop_reference.py
```diff
@@ -56,7 +56,11 @@ async def _run_prop_reference_asset(
     manager = get_task_manager()
     egress_context = _image_egress_context(envelope)
 
-    store = CogneeStore(ctx.owner_project_label, output_dir=str(output_dir))
+    store = CogneeStore(
+        ctx.owner_project_label,
+        output_dir=str(output_dir),
+        state_dir=str(ctx.state_dir),
+    )
```
lywaterman, line 63: **medium — the store is fixed here, but the style lookup at line 86 is not.**

The PR description lists "prop reference assets" as an affected path, yet `generate_prop_reference(..., project_dir=str(output_dir), ...)` at line 86 reaches `nanobanana_prop.py:78` → `get_style_preset(style, project_dir=...)` → `StyleService._resolve_project_context` → `Path(project_dir).relative_to(OUTPUT_DIR).parts[0:2]`.

For a scoped project whose `output_dir` is e.g. `output/_orgs/<org>/alice/demo`, that yields `username="_orgs", project="<org>"`. The custom-style map comes back empty and `get_style_preset` raises `KeyError: Style '<id>' not found` — the task fails outright, it does not silently fall back.

`scene_reference.py:102` was fixed for exactly this by threading `state_dir=str(ctx.state_dir)` through `get_style_preset`. The same one-line change is missing here.

src/novelvideo/task_backend/runners/character_image.py
```diff
@@ -129,14 +129,18 @@ def update(progress: float, current_task: str) -> None:
         )
 
     update(0.10, "加载角色数据...")
-    store = CogneeStore(ctx.owner_project_label, output_dir=str(output_dir))
+    store = CogneeStore(
+        ctx.owner_project_label,
+        output_dir=str(output_dir),
+        state_dir=str(ctx.state_dir),
```
lywaterman, line 135: **medium — same scoped-style gap as `prop_reference.py`, at lines 233, 285 and 380.**

The runner now correctly reads `ethnicity` from `ctx.state_dir`, but then passes `project_dir=str(output_dir)` into `generate_character_reference_unified`, which resolves custom styles via `nanobanana_character.py:256/475/697` → `get_style_preset(style, project_dir=...)` → the same `relative_to(OUTPUT_DIR).parts[0:2]` inference.

Net effect: a scoped project using a custom style fails with `KeyError` on portrait/identity generation while the equivalent scene-reference task succeeds — inconsistent behavior across runners for the same project. Threading `state_dir` through the same way `scene_reference.py:102` now does would close it.

src/novelvideo/services/style_service.py
```diff
@@ -172,7 +176,12 @@ def _load_project_custom_style_map(
         username: str | None = None,
         project: str | None = None,
         project_dir: str | Path | None = None,
+        state_dir: str | Path | None = None,
     ) -> dict[str, dict]:
+        if state_dir is not None:
+            config = load_project_config_file_from_state_dir(state_dir)
+            styles = config.get("custom_styles") or {}
+            return styles if isinstance(styles, dict) else {}
```
lywaterman, line 185: **medium — this makes the custom-style read path `state_dir`-keyed while the write path stays `owner/project`-keyed.**

The new short-circuit reads `custom_styles` from `<state_dir>/project_config.json`. But every writer still goes through `_save_project_custom_style_map` → `update_project_config_file(username, project, ...)` → `get_project_config_path(username, project)` = `STATE_DIR/<owner>/<project>/project_config.json`, and that function takes no `state_dir`.

For a default-layout personal project the two paths are the same file, so nothing changes. For a **scoped** project — the case this PR exists for — they diverge:

- `POST /styles` (`api/routes/styles.py:182`) writes to `STATE_DIR/<owner>/<project>/project_config.json`
- `GET /styles` and `GET /styles/{id}` (lines 70, 94) read from that same personal path
- `scene_reference` now reads from `ctx.state_dir`

So the UI shows and saves a custom style that the scene-reference runner can no longer find, and it falls back to the system preset.

`resolve_project_scope` already returns `state_dir` (`api/deps.py:118`), so the styles routes have the authoritative path in hand — threading it through `save_custom_style` / `_save_project_custom_style_map` / `list_custom_styles` would make read and write agree again. Worth doing in this PR since the read side just moved.

src/novelvideo/sqlite_store.py
```diff
@@ -339,7 +339,9 @@ def __init__(
...
+            default_state_dir = Path(state_dir)
```
lywaterman: **low — dead assignment, and the control flow is now split across two `if state_dir` blocks.**

`default_state_dir = Path(state_dir)` is never read: twelve lines below, `resolved_state_dir = Path(state_dir)` overrides it unconditionally whenever `state_dir` is truthy, and `default_state_dir` is only consulted in the `else` branch. The same dead assignment exists at `cognee/store.py:148`.

It also forces `state_dir` handling into two separate blocks (341-345 and 353-358), which obscures exactly the control flow this PR is correcting. A single block would be equivalent and easier to read:

```python
if state_dir:
    resolved_state_dir = Path(state_dir)
    if len(parts) == 2:
        from novelvideo.utils.project_paths import ProjectPaths
        paths = ProjectPaths(parts[0], parts[1])
        if resolved_state_dir.resolve() == paths.state_dir.resolve():
            paths.bootstrap_from_legacy_output()
elif len(parts) == 2:
    ...
```

src/novelvideo/cognee/store.py
```diff
@@ -144,7 +144,16 @@ def __init__(
...
+            default_state_dir = Path(state_dir)
+            if not sqlite_state_dir and "/" in project_name:
```
lywaterman: **low — this bootstrap block is unreachable in production.**

It only runs when `sqlite_state_dir` is empty, i.e. no injected store. But in exactly that case `__init__` immediately constructs `SQLiteStore(project_name, output_dir=..., state_dir=self.state_dir)` a few lines below, and `SQLiteStore.__init__` performs the identical `ProjectPaths` comparison and `bootstrap_from_legacy_output()` call. Nothing between the two points reads state-dir contents, so the migration always happens either way.

Worth noting that `test_cognee_store_explicit_personal_state_dir_migrates_legacy_config` only exercises this branch because it monkeypatches `SQLiteStore` out with a no-op fake — the block is covered by a scenario that cannot occur at runtime.

Suggest deleting it and letting `SQLiteStore` own the migration; that also removes the duplicated `ProjectPaths` import and comparison.

lywaterman (commented):
## Re-review of `c65f67d2`

**All five findings from the previous round are correctly addressed, and I verified the legacy-migration fix actually works.** Registry records store `str((Path(config.STATE_DIR)/user/project).resolve())` (`src/novelvideo/shared/project_dirs.py:13`), so the new `Path(state_dir).resolve() == paths.state_dir.resolve()` equality check does fire for existing personal projects and `bootstrap_from_legacy_output()` is preserved, while genuinely scope-owned paths still skip the derived fallback. `script.py` now uses `load_project_config_from_state_dir(...)`, keeping the effective-config layer and the `project_uuid` backfill. The PR's own tests plus the related `style` / `project_config` / runner / store suites pass locally (238 + 39 tests).

The store wiring itself I now consider done. What is left is that **the scope-awareness fix is applied to only one of the runners that needs it**, plus a resource leak the PR makes materially worse.

| # | File | Severity |
|---|------|----------|
| 1 | `runners/sketch.py:169` | medium — store never closed; leak now holds the *live* project DB open |
| 2 | `runners/prop_reference.py:86` | medium — scoped custom style raises `KeyError`, fix applied to `scene_reference` only |
| 3 | `runners/character_image.py:233` (also 285, 380) | medium — same gap |
| 4 | `services/style_service.py:181` | medium — read side is now `state_dir`-keyed, write side is not |
| 5 | `sqlite_store.py:343` / `cognee/store.py:148` | low — dead assignment, split control flow |
| 6 | `cognee/store.py:149` | low — bootstrap block is unreachable in production |

Findings 2-4 are all the same theme: `project_config.json` for a scoped project lives under `ctx.state_dir`, but the custom-style read and write paths still key on `owner/project` or on `project_dir` inference. `scene_reference` was fixed; the sibling runners and the styles API were not, so a scoped project that uses a custom style now behaves differently depending on which task you run.

Details inline.

Handanhhhy:
@lywaterman The latest review findings are addressed in `1c978b1a` and `4eddad88`. The follow-up commit also closes three propagation gaps caught by the pre-push audit: cleanup now covers sketch store initialization/loading failures, scoped style metadata reaches the final NanoBanana portrait/identity prompt branch, and the Seedream fallback receives the explicit state directory. Verification: 3213 passed / 16 skipped in the split full suite, the isolated order-dependent test passed, Ruff passed, and `git diff --check` passed. Could you please re-review when convenient?

src/novelvideo/services/style_service.py
```diff
@@ -188,15 +198,19 @@ def _save_project_custom_style_map(
         username: str | None = None,
```
lywaterman, line 198: **medium — this write path no longer triggers legacy migration, and can permanently orphan the rest of `project_config.json`.**

`update_project_config_file_in_state_dir` writes straight to `<state_dir>/project_config.json`. The path it replaces went through `update_project_config_file` → `get_project_config_path`, which calls `ProjectPaths.bootstrap_from_legacy_output()` first (`project_config.py:55`).

Scenario — a project still on the v1 single-directory layout (config only at `output/<user>/<project>/project_config.json`, no `state/<user>/<project>/.migrated`):

1. `POST /styles` creates `state/<user>/<project>/project_config.json` containing **only** `custom_styles`.
2. Later a store is opened and bootstrap finally runs. `_copy_missing_items` (`utils/project_paths.py:191`) skips `project_config.json` because `dst.exists()`, then touches the migration marker.
3. `ethnicity`, `visual_style`, `spine_template`, `project_uuid` and the narrator-reference keys are now permanently orphaned in the legacy file — the marker means bootstrap will never retry.

Unlike the five runners, the styles routes open no store first, so nothing bootstraps ahead of this write. Suggest running the same conditional bootstrap here that `SQLiteStore.__init__` now does (compare against `ProjectPaths(owner, project).state_dir`) before the first write, or having `update_project_config_file_in_state_dir` merge from the legacy file when the marker is absent.

src/novelvideo/services/style_service.py
```diff
@@ -172,7 +177,12 @@ def _load_project_custom_style_map(
...
         project: str | None = None,
         project_dir: str | Path | None = None,
+        state_dir: str | Path | None = None,
     ) -> dict[str, dict]:
```
lywaterman, line 181: **low — the read short-circuit ignores `username`/`project` entirely, so pre-PR custom styles are invisible.**

Once `state_dir is not None` this returns whatever is in `<state_dir>/project_config.json` and never consults `_resolve_project_context`.

Every write before this PR landed in `STATE_DIR/<owner>/<project>/project_config.json`, because `_resolve_project_context` returned the owner label regardless of scope. For a scoped project those existing custom styles are not migrated to the scoped `state_dir`, so after deploy:

- `GET /styles?project=<scoped>` silently returns presets only
- any project whose `visual_style` names one of those styles falls back to the default preset in `_effective_project_config`

The styles just vanish from the UI rather than erroring. A one-time read-through fallback (try `state_dir`, then the owner-derived path) or a migration step would avoid the silent loss.

src/novelvideo/generators/image_generator.py
```diff
@@ -995,7 +996,11 @@ async def generate_character_reference(
         os.makedirs(output_dir, exist_ok=True)
 
         style = style or self.default_style
```
lywaterman, line 998: **low — the Seedream branch resolves the style twice, and the second lookup is still scope-blind.**

`generate_character_reference` now resolves the preset with `state_dir` here, but then calls `self.generate(..., project_dir=project_dir)`. `ImageGenParams` has no `state_dir` field, so `_call_seedream_api` (line 362) re-resolves with `get_style_preset(params.style, project_dir=params.project_dir)`.

For a project whose `state_dir` is not `STATE_DIR/<owner>/<project>`, the first lookup now succeeds while the second raises `KeyError: Style '<id>' not found`. `generate()`'s bare `except Exception` converts that to `success=False`, and the runner reports the generic `角色 Portrait 生成失败` with the real cause erased.

Before this PR both lookups failed identically and the `KeyError` propagated with its message intact — so this change turns a loud failure into a silent one, which is the harder direction to debug. Mitigated in practice only if every scoped project is also an org tenant, where `ORG_EGRESS_DENIED` blocks seedream first. Adding `state_dir` to `ImageGenParams` would close it.

src/novelvideo/task_backend/runners/character_image.py
```diff
@@ -129,21 +129,26 @@ def update(progress: float, current_task: str) -> None:
...
+    )
     await store.initialize()
```
lywaterman, line 137: **low — `initialize()` sits outside the `try` whose `finally` closes the store.**

```python
store = CogneeStore(...)
await store.initialize()          # <- outside
await store.load_graph_state()    # <- outside
try:
    ...
finally:
    await store.close()
```

`CogneeStore.initialize` opens the aiosqlite connection via `_ensure_db()` *before* calling cognee `setup()`, so a `setup()` failure leaks the connection and file handle for the worker's lifetime.

Same shape at `prop_reference.py:57` and `scene_reference.py:81`. Pre-existing, but this PR fixed exactly this pattern in `sketch.py` (`store = None` + `try` wrapping construction and init) while touching all three of these constructor sites — worth applying the same shape here for consistency.

lywaterman (commented):
## Re-review of `1c978b1a` + `4eddad88`

**All six findings from the last round are correctly addressed**, and I verified the propagation is complete rather than spot-fixed:

- every changed generator guards with `state_dir or None` before hitting `StyleService` — I checked each call site, none is missed
- no caller passes the newly-inserted parameters positionally (`freezone/presets.py` uses keywords only), so widening those signatures is safe
- all 10 `CogneeStore(...)` sites in `task_backend/runners/` now pass `state_dir`, and `MockImageGenerator` is never handed the new kwarg
- removing `bootstrap_from_legacy_output()` from `CogneeStore.__init__` is safe: `SQLiteStore.__init__` is always constructed underneath with the same `state_dir` and still runs it when the dir equals `ProjectPaths.state_dir`
- the `sketch.py` `try/finally` is a genuine leak fix, and it correctly covers `initialize()` / `load_graph_state()` failures via the `store = None` guard
- the styles routes now thread `resolved.state_dir` through list / get / save / delete / preview, so the read-write split I raised last round is closed

### Findings

| # | Location | Severity |
|---|----------|----------|
| 1 | `style_service.py:198` | medium — the new write path skips legacy bootstrap and can permanently orphan the rest of `project_config.json` |
| 2 | `style_service.py:181` | low — read side ignores `username`/`project` entirely; pre-PR custom styles are not migrated |
| 3 | `image_generator.py:998` | low — Seedream resolves the style twice; the second lookup turns a loud `KeyError` into a silent failure |
| 4 | `character_image.py:132` (also `prop_reference.py:57`, `scene_reference.py:81`) | low — `initialize()` outside the `try`, the exact pattern `sketch.py` just fixed |

Finding 1 is the one worth acting on before merge — the styles routes are now the only writers of `project_config.json` that open no store first, so nothing bootstraps ahead of them.

Details inline.

lywaterman (approved):

PR #310, opened 2026-08-18
Fix typed scene task-start errors

Handanhhhy:
## Summary

- preserve typed billing, authorization, and task-limit failures across all three scene task-start endpoints while keeping ordinary RuntimeError compatibility responses
- centralize the task-start RuntimeError boundary and keep Freezone billing subtype priority intact
- add a typed, redacted FeatureBillingIdentityError contract for EE Release B
- publish and enforce the seven-step stage_asset contract before task-manager side effects, with real dispatch drift coverage

## Root cause

Several scene routes caught RuntimeError broadly. Because billing and authorization failures inherit from RuntimeError, those routes converted structured 402, 409, and 429 failures into legacy HTTP 200 error bodies. The stage runner also exposed no enforceable shared step contract for EE billing policy.

## Carried-forward review feedback

- [x] preserve wrapped task-limit failures and existing 429 handlers
- [x] retain InsufficientCredits then BillingRuleNotConfigured then generic BillingError priority
- [x] cover single-face 3GS, pano-to-3GS, and panorama generation task-start routes
- [x] provide a stable redacted 409 billing-identity error
- [x] validate unknown stage steps before manager, progress, cancellation, or task-function side effects
- [x] execute all seven legal dispatch branches in the CE drift guard
- [x] verify existing acknowledge-no-refund cancellation behavior

## Validation

- backend focused and adjacent regression suite: 296 passed
- frontend paid-task cancellation suite: 5 passed
- Ruff passed for all changed files
- Black passed for the seven scoped files already formatted by staging; the current staging version of freezone.py has existing whole-file Black debt
- git diff check passed
- DCO sign-off present

## Rollout

This CE repair must be merged and deployed before EE Release B activates operation-level stage_asset charging. EE Release A only registers the four price keys and remains behavior-neutral.

### DIFF
--- a/src/novelvideo/api/routes/scenes.py
+++ b/src/novelvideo/api/routes/scenes.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import io
+import logging
 import shutil
 import tempfile
 import time
@@ -24,6 +25,7 @@ from novelvideo.api.schemas import (
     SceneReferenceGenerateRequest,
     SceneUpdate,
 )
+from novelvideo.api.task_start_errors import handle_task_start_runtime_error
 from novelvideo.api.viewer_manifests import (
     build_director_stage_manifest,
     build_pano_viewer_manifest,
@@ -51,6 +53,7 @@ from novelvideo.utils.path_resolver import (
 )
 
 router = APIRouter()
+logger = logging.getLogger("novelvideo.api.scenes")
 
 _SCENE_TIME_TOKENS = {
     "清晨",
@@ -1404,6 +1407,11 @@ async def _start_3gs_single_face_task(
             params=params,
         )
     except RuntimeError as exc:
+        handle_task_start_runtime_error(
+            logger,
+            "failed to start single-face stage asset task",
+            exc,
+        )
         return {"ok": False, "error": str(exc)}
 
     return {
@@ -1489,6 +1497,11 @@ async def generate_scene_3gs_pano_ply(
             params=params,
         )
     except RuntimeError as exc:
+        handle_task_start_runtime_error(
+            logger,
+            "failed to start pano stage asset task",
+            exc,
+        )
         return {"ok": False, "error": str(exc)}
 
     return {
@@ -1539,6 +1552,11 @@ async def generate_scene_pano(
             params=params,
         )
     except RuntimeError as exc:
+        handle_task_start_runtime_error(
+            logger,
+            "failed to start scene pano generation task",
+            exc,
+        )
         return {"ok": False, "error": str(exc)}
 
     return {
--- a/src/novelvideo/shared/billing_errors.py
+++ b/src/novelvideo/shared/billing_errors.py
@@ -8,6 +8,9 @@ INSUFFICIENT_CREDITS_CODE = "INSUFFICIENT_CREDITS"
 INSUFFICIENT_CREDITS_MESSAGE = "积分不足，请联系管理员充值"
 BILLING_RULE_NOT_CONFIGURED_CODE = "BILLING_RULE_NOT_CONFIGURED"
 BILLING_RULE_NOT_CONFIGURED_MESSAGE = "计费规则未配置，请联系管理员设置积分规则"
+FEATURE_BILLING_IDENTITY_INVALID_CODE = "FEATURE_BILLING_IDENTITY_INVALID"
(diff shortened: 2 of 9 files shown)


src/novelvideo/shared/billing_errors.py
```diff
@@ -76,6 +79,22 @@ def details(self) -> dict[str, Any]:
...
 
+class FeatureBillingIdentityError(BillingError):
+    """Raised when the billing identity for a feature cannot be trusted."""
+
+    error_code = FEATURE_BILLING_IDENTITY_INVALID_CODE
+    http_status = 409
+    user_message = FEATURE_BILLING_IDENTITY_INVALID_MESSAGE
+
+    def __init__(self, *, internal_reason: str = "") -> None:
+        safe_reason = internal_reason if type(internal_reason) is str else ""
```
lywaterman, line 90: **nit** — `type(internal_reason) is str` rejects `str` subclasses, where `isinstance(internal_reason, str)` would accept them. Nothing in the tree passes one today, so this is style rather than a defect; `isinstance` is just the conventional spelling and avoids a surprise if a caller ever passes something like an `enum.StrEnum` member.

Also worth noting for whoever wires this up in EE: passing a non-`str` silently degrades to the generic message rather than failing loudly, which could hide a wiring mistake at the call site.

src/novelvideo/api/routes/scenes.py
```diff
@@ -1404,6 +1407,11 @@ async def _start_3gs_single_face_task(
             params=params,
         )
     except RuntimeError as exc:
+        handle_task_start_runtime_error(
+            logger,
+            "failed to start single-face stage asset task",
+            exc,
+        )
         return {"ok": False, "error": str(exc)}
```
lywaterman, line 1415: **nit** — for RuntimeErrors that are *not* typed, this still returns the raw `str(exc)` to the client.

That is pre-existing behavior and out of scope here, but it sits a little oddly next to `FeatureBillingIdentityError`, which this same PR adds specifically to keep internal reasons out of the response body. If the redaction posture is deliberate, these three sites are the natural follow-up — a generic message to the client plus `logger.warning` for the detail, which the helper already does.

lywaterman (commented):
## Review

**No blocking findings.** I verified the parts that are easy to get wrong rather than taking the PR description at its word:

**Coverage is complete.** I scanned every route file on `staging` for the `except RuntimeError` → `return {"ok": False, ...}` shape. It occurs in exactly three places, all in `scenes.py`, and all three are fixed here — which covers four endpoints, since `_start_3gs_single_face_task` serves both `3gs/master-ply/generate-async` and `3gs/reverse-ply/generate-async`. The six `HTTPException(503)` task-start sites in `freezone.py` already routed through `_handle_task_start_runtime_error`, so they inherit the shared helper for free.

**`SUPPORTED_STAGE_ASSET_STEPS` matches the dispatch exactly** — seven frozenset entries against seven `if/elif` branches (`splat_collision`, `upload_scene_package`, `pano_sharp`, `single_face_sharp`, `voxel_world_from_360`, `pano_from_master`, `pano_from_text`), nothing extra, nothing missing. The guard sits before `get_task_manager()` and before the first `update()`, so it genuinely precedes the manager, progress and cancellation side effects. The trailing `raise ValueError` at line 220 is not dead code: it still catches the reverse drift, where the frozenset gains an entry the dispatch doesn't handle.

**The redaction actually holds.** `app.py`'s `BillingError` handler renders `exc.user_message` plus `billing_error_payload(exc)`, and that payload only reads `error_code` / `user_message` / `details()`. `FeatureBillingIdentityError` does not override `details()`, so `internal_reason` reaches the log and never the response body.

**No orphaned references** in `freezone.py` after the import block was removed — `find_authz_error` is still imported and still used at line 7820.

**The typed errors survive the trip to the UI.** I checked whether converting these three routes from `200 {ok:false}` to a typed 4xx would lose the message: `frontend/src/api/client.ts` has a ky `beforeError` hook that parses the response body, extracts `error`, builds an `ApiError` (or a typed one via `errorFromBackendBody`) and attaches it as `.cause`, which `apiCall` re-raises. So the localized `user_message` and the machine `error_code` both reach the caller — no regression there.

Two nits inline, neither worth blocking on.

lywaterman (approved):

Issue #76, opened 2026-07-08
Fix episode prop planner schema failures and prior-prop reuse


Handanhhhy:
## Problem
Episode prop planning could fail after import even when scenes and earlier prop blocks were valid.

Observed cases:
- The model returned `requirements` as a string list instead of an object array, causing `BlockPropRequirements` validation to fail with `Exceeded maximum output retries`.
- In精品剧/test04, a later short scene block only reused an already selected prop (`羊皮笔记本` -> `笔记本`), but the planner still called the model and failed the whole episode. Earlier valid props were not persisted because episode prop planning writes all-or-nothing.

## Expected
- Invalid structured output should remain a failure, with a clear diagnostic.
- Short blocks that only mention already selected episode props should deterministically reuse those props without calling the model again.
- No arbitrary fallback should create new props from invalid model output.

## Verification
- Docker test04 latest `episode_prop_planner` runs completed and wrote prop menus.
- `uv run pytest tests/test_format_check.py tests/test_asset_compiler_prop_planning.py tests/test_newapi_text_gateway.py::test_asset_compiler_prop_planner_uses_prop_newapi_env tests/test_asset_compiler_scene_enrichment.py`
- `uv run ruff check src/novelvideo/agents/asset_compiler.py tests/test_asset_compiler_prop_planning.py`

PR #311, opened 2026-08-18
fix authz transient retry handling

Handanhhhy:
## Summary

- classify authorization infrastructure failures separately from deterministic authorization drift
- add bounded full-jitter retries at request, inline-worker, and post-provider revalidation boundaries
- emit stable local retry scheduled/recovered/exhausted events at each bounded retry call site
- preserve the single refund path and send post-provider uncertainty to review without replaying the provider
- close review findings around exception-chain redaction, post-accept gray disable, safe refund diagnostics, and the unreachable inline pseudo-refund path

## Root cause

Database and authorization-service failures were flattened into deterministic stale-envelope failures. That could incorrectly fail and refund queued work, while later revalidation boundaries could hard-fail running tasks without preserving settlement uncertainty.

## Impact

Transient authorization outages are now retried safely. Genuine authority drift remains terminal, running tasks are never automatically refunded after provider acceptance, and typed failures remain visible to the EE Celery layer. `P0_GRAY_DISABLED` remains an operator interruption and uses the existing failure/refund path after provider acceptance rather than being misclassified as authority drift.

## CE/EE deployment coupling

This PR must deploy with [SuperTale PR #66](https://github.com/claymorelab/SuperTale/pull/66). Mixed old/new CE and EE API/worker combinations are unsupported because the rejected-settlement shape and terminalization contract change together.

Release requirements:

1. Build with the exact CE and EE refs from these two PRs.
2. Confirm PostgreSQL 16+ and apply EE migration 0078.
3. Start the isolated `task.settlement` worker.
4. Verify via Celery `inspect active_queues` that at least one live worker consumes `task.settlement`; the general worker intentionally does not.
5. Roll API and workers together and monitor retry exhaustion, terminalization faults, review age, and 0077 reaper recovery.

## Review closure

- CE inline does not create feature-credit reservations, so the unreachable pseudo-refund and artificial metadata-injection tests were removed; EE registration is ratcheted to Celery for control-plane billing.
- Typed authz failures have their cause/context/old traceback removed before crossing producer and inline boundaries.
- Inline authorization retry now reuses the shared helper and emits the common retry events.
- Refund failures log an opaque error id and safe exception type without logging raw driver text or traceback.
- Generic post-accept exceptions retain a secret-canary redaction test.

The normal runner's existing outer billing-metadata trust boundary is tracked separately; this review closure does not introduce a second signing protocol or an Outbox lookup into the runner.

## Validation

- CE focused authz/task/egress regression suite: `103 passed`
- Companion EE focused settlement/terminalizer suite: `59 passed`
- Black and Ruff passed on changed CE Python files
- `git diff --check` passed in both repositories


src/novelvideo/ports/local/tasks.py
```diff
@@ -275,25 +290,39 @@ async def _run_inline(
...
+                    )
+                    await _AUTHZ_RETRY_SLEEP(_AUTHZ_RETRY_RANDOM() * upper_bound)
         if consumer_failure is not None:
             settlement = consumer_failure.settlement
             if settlement is not None:
                 # 预留发生在入队侧、退款发生在 run_core 里,而这里是两者之间。
                 # 只有签名已验证的拒绝才带 settlement——未验证的信封不驱动资金。
                 await refund_undelivered_feature_credit_reservation(
-                    feature_credit_reservation_id(settlement.billing_metadata),
+                    feature_credit_reservation_id(
```
lywaterman: **high — this refund can never fire, so a verified-then-rejected delivery silently keeps the reservation.**

`job.metadata` is built in `enqueue_project_task` (line 106) as:

```python
metadata = {
    "backend": "inline",
    "queue_kind": lane_name,
    "project_id": ctx.project_id,
    "product_surface": product_surface,
    **display_metadata_for_task(task_type, payload),
}
```

`display_metadata_for_task` (`ports/tasks.py:17`) only ever returns a whitelist of display strings — `display_name`, `task_label`, `task_family`, `source_label`, `target_label`, `canvas_id`, `node_id`, `skill_id`, plus `scene_name`/`step` for `stage_asset`. No production path writes `billing_metadata` into this dict; the only writers are the tests (`job.metadata["billing_metadata"] = ...`).

So `feature_credit_reservation_id(...)` always gets `{}` → returns `""` → `refund_undelivered_feature_credit_reservation` hits its `if not reservation_id: return accepted=True` short-circuit. A stale-authority or retry-exhausted rejection on the inline lane fails the task and leaves the credit row in `feature_reserved` forever — the exact OI-59 leak the comment above says this block exists to close. The tests pass because they inject the key the runtime never sets.

There is also a threat-model inconsistency here. The rationale for moving off the delivery is that the envelope is attacker-controlled — but `run_core.run_project_task_core_sync` still derives `feature_reservation_id` from `delivery.billing_metadata`, i.e. `raw_delivery["billing_metadata"]` (`run_core.py:613`, `:630`), so the same untrusted value still drives every refund and confirm on the main path. Either the threat model is wrong and this should read from the delivery like run_core does, or run_core is still exposed and needs the same treatment. Right now it is the worst of both: the hardened path is inert and the unhardened path is unchanged.

src/novelvideo/generators/video_generator.py
```diff
@@ -2777,6 +2795,12 @@ async def _revalidate_organization(self, context=None) -> None:
         ):
             raise AuthzError("ORG_AUTHZ_STALE")
 
+    @staticmethod
+    def _running_authority_error(exc: AuthzError) -> RunningTaskAuthorityIndeterminate:
```
lywaterman: **medium — `P0_GRAY_DISABLED` is an operator toggle, not authority drift, and it now parks money in "review".**

`_running_authority_error` defaults `failure_kind` to `"drift"` for any plain `AuthzError`, and `run_core` routes every `RunningTaskAuthorityIndeterminate` to `mark_feature_credit_settlement_for_review` with no refund. `admit_model_task` raises `P0_GRAY_DISABLED` from its very first `require_enabled()` line, so it lands in that bucket.

Scenario: an operator flips `ST_P0_GRAY_ENABLED` off while org video tasks are polling. Every in-flight task now fails with `TASK_AUTHZ_REVALIDATION_INDETERMINATE` and its feature-credit reservation is parked in review instead of refunded — previously the runner raised `RuntimeError(result.error)` and `run_core` refunded it.

In CE this is worse than a hold: `ports/local/usage.py`'s `mark_feature_credit_settlement_for_review` is a stub returning `{"status": "awaiting"}` that persists nothing. The reservation is orphaned with no durable record that anyone needs to look at it.

Suggest excluding `P0_GRAY_DISABLED` — and generally any code that is not a change in *user* authority — from the review path, and letting it refund as before.

src/novelvideo/task_backend/producer.py
```diff
@@ -121,21 +121,21 @@ async def sign_top_level(
...
-            authz_error_code = exc.code
+            authz_failure = exc
             admission = None
         except Exception:
             ordinary_failure = True
             admission = None
-        if authz_error_code is not None:
-            raise AuthzError(authz_error_code) from None
+        if authz_failure is not None:
+            raise authz_failure from None
```
lywaterman: **medium — `raise ... from None` does not clear `__context__`, so the raw EE PostgresError text is still reachable.**

This site (and `ports/local/tasks.py:190`) changed from constructing a fresh `AuthzError(code)` to re-raising the caught EE exception object. The old shape was deliberate — `egress_binding.py` spells out why: EE's `authz/admission_repository.py` maps any `PostgresError` to `ORG_AUTHZ_STALE`, and 原始 PG 错误文本会挂在异常上下文里随之外泄.

`raise x from None` only sets `__cause__ = None` and `__suppress_context__ = True`. It does **not** clear `__context__`, and `shared/billing_errors.iter_exception_chain` walks

```python
current = current.__cause__ or current.__context__
```

unconditionally, ignoring `__suppress_context__`. So `find_authz_error` / `find_billing_error` — and any error reporter that walks the chain — still reach the EE-internal DB text.

`egress_binding.py:95-97` handles this correctly:

```python
authz_failure.__cause__ = None
authz_failure.__context__ = None
authz_failure.__traceback__ = None
raise authz_failure from None
```

These two sites need the same clearing, or should keep reconstructing a fresh error.

src/novelvideo/task_backend/run_core.py
```diff
@@ -333,17 +347,25 @@ async def refund_undelivered_feature_credit_reservation(
...
         )
-    except Exception as exc:  # noqa: BLE001
+    except FeatureCreditSettlementConflict:
+        logger.warning(
+            "undelivered feature credit refund conflicts with durable settlement",
+            extra={"failure_kind": "settlement_action_conflict"},
+        )
+        return SettlementIntentResult(accepted=False, retryable=False)
+    except Exception:  # noqa: BLE001
         logger.error(
```
lywaterman, line 433: **low — the refund-failure log lost its payload.**

This was `logger.error("...: %s", exc)`; it is now a constant string plus `extra={"failure_kind": "settlement_adapter_failure"}`, with no `exc` and no `exc_info=True`.

This is the money path. When a refund fails, whoever is paged gets a fixed sentence and no cause — and the new `SettlementIntentResult` return value is discarded by every caller, so nothing downstream records it either. Adding `exc_info=True` keeps the structured `extra` while restoring the diagnosis.

src/novelvideo/ports/local/tasks.py
```diff
@@ -275,25 +290,39 @@ async def _run_inline(
         lane: _InlineLane,
         job: _InlineLaneJob,
     ) -> None:
-        consumer_failure: InvalidTaskEnvelope | None = None
+        consumer_failure: InvalidTaskEnvelope | TaskAuthorityUnavailable | None = None
         verified = None
         if self._consumer is None:
             consumer_failure = InvalidTaskEnvelope()
         else:
```
lywaterman, line 297: **low — this is a hand-rolled copy of `retry_authz_read` rather than a call to it.**

Consequences: it skips `validate_authz_retry_policy`, and it emits none of the `authz_local_retry_scheduled` / `_recovered` / `_exhausted` events the PR description promises "at each bounded retry call site". The inline boundary is the one that can burn ~62 s of wall clock per job, so it is the site where that observability matters most — and it is the only one without it.

Related: `retry_authz_read`'s `deadline` and `check_cancelled` parameters are dead — no call site passes either. Worth either wiring them here or dropping them until there is a caller.

tests/test_p0g4c_video_egress.py
```diff
@@ -739,9 +740,7 @@ async def post(*_args, **_kwargs):
         return {"id": "provider-job-1"}
 
     async def disabled(_context):
-        error = RuntimeError("disabled detail canary")
-        error.code = disabled_code
-        raise error
+        raise AuthzError(disabled_code)
```
lywaterman: **low — this rewrite removes the only coverage of the generic branch's redaction.**

These three canaries (also `:793`, `:909`) now raise a real `AuthzError`, which takes the new `except AuthzError` path. That leaves `_safe_video_error_code(exc, "ORG_AUTHZ_STALE")` on the generic-exception branch at the post-acceptance revalidation sites with no test behind it.

Given that branch exists specifically to redact, a future regression that leaks exception text through it would now pass CI. Suggest keeping one case that raises a non-`AuthzError` exception.

lywaterman (commented):
## Review — CE side of the authz transient-retry pair

Reviewed together with EE claymorelab/SuperTale#66, since the two are a single expand/contract change and the interesting failure modes live in the seam.

I read every source hunk with surrounding context and ran the focused suite (`test_authz_retry.py`, `ports/test_tasks.py`, `ports/test_authz.py`, `test_task_envelope_consumer.py`, `test_task_envelope_producer.py`, `test_running_task_authz_indeterminate.py`, `test_undelivered_refund_result.py`, `test_request_egress_binding.py`, `test_authz_error_http_contract.py`) — **145 passed**. The three red CI checks are `dco` and `gitleaks`, not test failures.

**Cross-repo contract checks that pass:** `RejectedTaskSettlement` dropping `billing_metadata` and gaining `root_task_id` matches what EE's `task_terminalization.py` constructs (keyword args, so field order is not load-bearing). `FeatureCreditSettlementConflict` and `mark_feature_credit_settlement_for_review` are added to the `UsageMeter` Protocol and implemented on both sides. `refund_undelivered_feature_credit_reservation` returns `SettlementIntentResult(accepted, retryable)`, which is exactly what EE's `celery_tasks.py` consumes.

**Deploy coupling** — worth restating in the PR body, not just the EE design doc: EE `staging`'s `celery_tasks.py:101` reads `settlement.billing_metadata`, which this PR removes, and EE#66 constructs `RejectedTaskSettlement(root_task_id=...)`, which `staging` CE does not accept. Neither side can ship first. The design doc's §发布顺序 covers this correctly with expand/contract and an explicit ban on mixed combinations; the constraint just isn't visible from this PR alone.

### Findings

| # | Location | Severity |
|---|----------|----------|
| 1 | `ports/local/tasks.py:323` | **high** — the pre-worker refund can never fire; reservation is silently orphaned |
| 2 | `generators/video_generator.py:2799` | medium — `P0_GRAY_DISABLED` gets parked in "review" instead of refunded |
| 3 | `producer.py:138`, `ports/local/tasks.py:190` | medium — raw EE PostgresError text stays reachable on `__context__` |
| 4 | `run_core.py:363` | low — refund-failure log lost its cause |
| 5 | `ports/local/tasks.py:297` | low — hand-rolled retry, invisible in logs |
| 6 | `tests/test_p0g4c_video_egress.py:743` | low — lost the only redaction coverage for the generic branch |

Finding 1 is the blocker. Finding 3 matters because the whole point of `egress_binding.py:95-97` is that this text must not travel.

Also non-blocking: `license-inventory.csv` picks up 102 unrelated generated rows, which looks like a rebase artifact.

tests/test_p0g4c_video_egress.py
Handanhhhy: Addressed in be263502. A dedicated post-accept test now raises a generic RuntimeError containing a DSN secret canary and verifies the returned error is ORG_AUTHZ_STALE, the operation becomes unknown, and the secret is absent.

src/novelvideo/ports/local/tasks.py
Handanhhhy: Addressed in be263502 using the verified current product invariant. CE inline creates no feature-credit reservation, while EE control-plane registration always binds CeleryTaskBackend. The inert pseudo-refund and artificial billing_metadata injection tests were removed, and a Celery-only registration ratchet was added. The normal runner outer billing-metadata trust boundary is explicitly tracked separately rather than introducing a second signing or lookup protocol in this closure.

src/novelvideo/generators/video_generator.py
Handanhhhy: Addressed in be263502. Post-provider-accept P0_GRAY_DISABLED is no longer converted to RunningTaskAuthorityIndeterminate: the operation is marked unknown and the existing failure/refund path handles it. Genuine membership/authz drift and unavailable/fault exhaustion remain review-only. Tests cover submit, poll/fetch, and last-frame boundaries.

src/novelvideo/task_backend/producer.py
Handanhhhy: Addressed in be263502. Producer and inline enqueue now call a shared detach_authz_error boundary helper that clears cause, context, and the old traceback before re-raising the same typed subtype. Regression tests construct an error carrying an internal Postgres-style canary chain and verify that it is removed.

src/novelvideo/task_backend/run_core.py
Handanhhhy: Addressed in be263502 with a redaction-safe diagnostic instead of exc_info=True. Refund-adapter failures now log failure_kind, safe_error_type, and an opaque error_id, while never logging the exception string or traceback because those can contain a DSN or SQL text. The test verifies both diagnosability fields and secret-canary absence.

src/novelvideo/ports/local/tasks.py
Handanhhhy: Addressed in be263502. Inline consumption now calls retry_authz_read with TaskAuthorityUnavailable as the explicit retryable boundary type, so it shares policy validation and scheduled/recovered/exhausted events. The generic deadline and cancellation hooks remain available for callers with a real deadline/cancellation source; this change does not invent one for inline execution.

Handanhhhy:
@lywaterman Review findings are addressed in be263502 and all six threads have replies with verification details. The PR body now spells out the CE/EE lockstep deployment, migration 0078, and isolated task.settlement worker gate. Focused CE suite: 103 passed; companion EE suite: 59 passed. Ready for re-review.

lywaterman (commented):
## Follow-up review — `0cd00082`

Re-reviewed at head `0cd00082` (`b0812d81 → b2e7c618 → 504db1e8 → 0cd00082`). Focused suite across 9 files (`ports/test_tasks.py`, `test_authz_retry.py`, consumer/producer, indeterminate, refund-result, egress-binding): **123 passed** in 30 s.

> **Blocking:** N1 below. This PR removed the unsigned-`billing_metadata` dependency from `RejectedTaskSettlement` (now `root_task_id`) but left it driving every confirm and refund on the main money path.

Two housekeeping notes before the findings: your comment says "all six threads have replies with verification details", but only one inline reply is visible on GitHub (the `test_p0g4c_video_egress.py` thread) — the rest may still be an unsubmitted draft. And the cited commit `be263502` is not on the branch (`git cat-file -t` fails); presumably a rebase changed the SHA.

### Verdicts on the 6 prior findings

**1. `ports/local/tasks.py:323` — pre-worker refund (high) — PARTIALLY FIXED.**

The dead-refund half is resolved: `0cd00082` deletes the whole `if settlement is not None: refund…` block from `_run_inline` along with the `feature_credit_reservation_id` / `refund_undelivered_feature_credit_reservation` imports. That is the right call for CE — `InlineTaskBackend.enqueue_project_task` (`:85-108`) has no `billing_metadata` parameter and never reserves anything, so CE inline has no reservation to orphan, and settlement responsibility moves to EE through the new signed `RejectedTaskSettlement.root_task_id`. `test_inline_stale_authority_never_initiates_feature_credit_settlement` codifies it.

The threat-model half is **not** fixed — see N1.

**2. `video_generator.py:2799` — `P0_GRAY_DISABLED` parked in review (medium) — FIXED.**

`_running_authority_error` now returns `detach_authz_error(exc)` for `P0_GRAY_DISABLED`, re-raising a plain `AuthzError`. It falls through `except RunningTaskAuthorityIndeterminate` / `VideoEgressError` / `NewApiVideoError` into the generic `except Exception` at `:3671`, which refunds the model-call reservation and returns `FAILED("P0_GRAY_DISABLED")` → the runner raises → `run_core` refunds the feature credit as before. Covered by `test_newapi_gray_disable_after_acceptance_uses_existing_failure_refund_path`, and refunding post-acceptance here is consistent with the existing timeout path at `:3598`.

**3. `producer.py:138` / `tasks.py:190` — `from None` leaves `__context__` (medium) — FIXED at both named sites, but reintroduced elsewhere.**

`detach_authz_error` was added at `ports/authz.py:107` and applied at both sites — both raise outside the `except` block, so nothing re-chains — with a canary test asserting `__cause__` / `__context__` / traceback are clean. The new `_running_authority_error` non-P0 branch reintroduces exactly this leak; see N2, which I reproduced.

**4. `run_core.py:363` — refund-failure log lost its cause (low) — PARTIALLY FIXED.**

`safe_error_type` restores the exception type, but there is still no `exc` and no `exc_info`. The added `error_id` is a fresh uuid that appears in no other record and is not returned in `SettlementIntentResult`, so it correlates with nothing. Both CE call sites still discard the return value, so `retryable` remains CE-dead (EE-only).

**5. `ports/local/tasks.py:297` — hand-rolled retry (low) — FIXED.**

Now calls `retry_authz_read(..., call_site="inline_task_consumer", retryable_error_types=(TaskAuthorityUnavailable,))`, which runs `validate_authz_retry_policy` and emits `authz_local_retry_scheduled` / `_recovered` / `_exhausted`, asserted in `test_inline_authz_retry_exhaustion_fails_exact_task_without_settlement`. `deadline` / `check_cancelled` are still dead — see N6.

**6. `tests/test_p0g4c_video_egress.py:743` — lost redaction coverage (low) — FIXED.**

`test_newapi_generic_post_accept_authz_failure_is_redacted` raises a bare `RuntimeError("postgres://user:secret-canary@internal")` from the post-accept revalidation and asserts `error == "ORG_AUTHZ_STALE"`, operation → `unknown`, canary absent.

**License-inventory noise — FIXED.** `git diff 7ca5efa8...HEAD -- license-inventory.csv` is now exactly 4 rows, all for files this PR adds.

### New findings

**N1. `src/novelvideo/task_backend/run_core.py:618` — medium (blocking) — confirm and refund are still keyed off an unsigned, delivery-supplied reservation id.**

`run_core.py:618` does `_clean_billing_metadata(delivery.billing_metadata)` → `run_metadata` → `feature_reservation_id` (`:634-635`), and `consumer.py:249-252` copies that straight out of `raw_delivery["billing_metadata"]`. `billing_metadata` is **not** in `_SIGNED_PAYLOAD_FIELDS` (`consumer.py:36`) and is not among the fields `SignedTaskEnvelope.verify` cross-checks, so it sits outside the HMAC.

Scenario: an actor who can write to the task queue replays their own validly signed envelope with `billing_metadata.feature_credit_reservation_id` set to another tenant's reservation. On completion `run_core.py:885` confirms — i.e. charges — that foreign reservation; on failure `:807` / `:829` refunds it.

**N2. `src/novelvideo/generators/video_generator.py:2804` — low — the non-`P0_GRAY_DISABLED` branch doesn't detach, so EE-internal text stays reachable.**

`raise self._running_authority_error(exc) from None` at `:3374` / `:3444` / `:3478` runs inside `except AuthzError as exc`, so the new `RunningTaskAuthorityIndeterminate` keeps `__context__ = exc`, and `exc.__context__` is whatever EE's `admission_repository` mapped to `ORG_AUTHZ_STALE`. Reproduced: `iter_exception_chain` yields `RunningTaskAuthorityIndeterminate → AuthzError → RuntimeError('postgres://user:SECRET@internal db error')`. The P0 branch two lines above detaches; this one doesn't. Call `detach_authz_error(exc)` before constructing, or clear `__context__` on the new error.

**N3. `src/novelvideo/task_backend/run_core.py:315` (and `:295`) — low/medium — `FeatureCreditSettlementConflict` is handled in only one of three settlement helpers.**

`_refund_feature_credit_reservation` (used at `:654` cancel-before-start, `:710`, `:726`) and `_confirm_feature_credit_reservation` (`:885`) still fall into `except Exception` and log "remains awaiting retry". When EE raises the conflict for a reservation that already advanced to a final action, ops get a permanently unresolvable "awaiting retry" alert instead of the non-retryable classification the sibling function now produces. These two also still log `"...: %s", exc` — the raw adapter text the sibling was just redacted to avoid.

**N4. `src/novelvideo/ports/local/tasks.py:296` — low — `fault` is now retried on the inline lane, holding a lane slot for up to ~62 s per task.**

The old loop retried only `failure_kind == "unavailable"`; the new call passes `retryable_error_types=(TaskAuthorityUnavailable,)`, and `consumer.py:224` maps *any* unexpected exception from `admit_model_task` (e.g. a `TypeError` in the EE adapter) to `fault`. So a deterministic bug now makes every inline job sleep through 5 backoffs (base 2 s, cap 60 s) before failing, while `lane.active` stays incremented (`_start_lane_job`) — the lane's whole concurrency budget stalls for a minute-plus per job instead of failing fast.

**N5. `src/novelvideo/generators/video_generator.py:3610` — low — on the indeterminate path the model-call reservation is neither refunded, confirmed, nor marked for review.**

`except RunningTaskAuthorityIndeterminate: raise` bypasses the outer handler's `_refund_video_model_call`, and `run_core`'s review marker covers only the *feature* credit. If authz is briefly unavailable during polling, the task fails with `TASK_AUTHZ_REVALIDATION_INDETERMINATE`, the feature credit goes to review, and the `reserve_current_model_call_credit` reservation from `:3307` stays in `reserved` with no durable record — while the neighbouring post-acceptance timeout path (`:3598`) does refund it with `provider_task_id` evidence.

**N6. `src/novelvideo/authz_retry.py:63-64` — low — `deadline` and `check_cancelled` remain dead parameters.**

No production call site passes either (grepped across `src/`), so the deadline branch at `:115` and both cancellation checks are unreachable outside `tests/test_authz_retry.py`, and the exhaustion log they emit can never fire in production.

Handanhhhy:
@lywaterman N1–N6 are now addressed at head 76399741.

Key changes:
- N1 strips delivery-controlled settlement IDs and resolves the feature reservation from the verified signed task identity.
- N2 moves transformed post-accept authz raises outside active exception handlers; the secret canary now proves the surfaced chain is clean.
- N3 handles all three feature settlement conflict paths consistently with redacted structured logs.
- N4 introduces a distinct fail-fast TaskAuthorityFault; only TaskAuthorityUnavailable is retried by inline and Celery paths.
- N5 sends accepted-provider model reservations to durable EE review with provider evidence, without automatic confirm/refund.
- N6 removes the dead deadline and cancellation callback parameters and unreachable tests.

Verification: CE 87 passed, 1 environment-only deselection; EE lockstep 65 passed, 1 skipped. Black, Ruff, and git diff --check passed. Please re-review together with claymorelab/SuperTale#66.

lywaterman (commented):
## Follow-up review 2 — `76399741`

Re-reviewed at head `76399741` (`73c4799f → 76399741`), alongside EE at SuperTale#66 head `8ac6f5fe`. Targeted run of `test_running_task_authz_indeterminate.py`, `test_trusted_task_settlement_resolution.py`, `ports/test_tasks.py`, `test_task_timeout.py`, `test_task_envelope_consumer.py`, `test_authz_retry.py`, `test_feature_settlement_safe_logging.py`, `test_p0g4c_video_egress.py`: **104 passed**, 0 failed, 0 deselected, ~4.7 s. I did not observe the "1 environment-only deselection" — presumably in a file I did not run.

**N1–N6 are all fixed.** I verified N1 by tracing the runtime path end to end rather than trusting the tests, and N2 by writing my own exception-chain reproduction rather than relying on the canary. Details below, then six new findings — the first two of which I'd ask you to resolve before merge.

### Verdicts

**N1 — FIXED.** `run_core.py:686` wraps `_clean_billing_metadata(delivery.billing_metadata)` in `_without_settlement_handles`, dropping `feature_credit_reservation_id` / `feature_credit_charge_id`; the same strip is applied to `run_metadata` (`:704`). `feature_reservation_id` now comes only from `_resolve_feature_reservation_id` (`:233`), whose `VerifiedTaskSettlementIdentity` is built from `delivery.admission.root_task_id`, `delivery.project_id`, `delivery.admission.requester_user_id`, `signed.task_type` and `signed_payload["episode"/"beat_num"/"scope"]` — every one HMAC-covered by `SignedTaskEnvelope.verify` (the first three are envelope fields; the last three are in `_SIGNED_PAYLOAD_FIELDS` and strictly compared), with `verify` also pinning `expected_root_task_id == run_task_id`.

EE's `ee_resolve_feature_settlement` (SuperTale migration 0078) additionally re-checks `states.task_id = p_expected_task_id AND states.requester_user_id = p_requester_user_id` and joins `credit_ledger.user_id` / `project_id`, so a replayed envelope cannot reach another tenant's reservation. Lookup-miss behaviour: no matching row/identity → `conflict` → the task fails before the runner and the reservation stays durably parked in EE's `review`/`awaiting` outbox row; no awaiting feature row → `not_applicable` → `""` and every settle helper short-circuits. No reader of `delivery.billing_metadata` remains on a money path (`:687` is the only one; `runners/episode_assets.py:60` reads the already-stripped envelope).

The strip is now too wide, though — see N7.

**N2 — FIXED, verified by independent reproduction.** I drove `NewApiVideoGenerator.generate` with `_revalidate_organization` raising `AuthzError` from inside an active `except RuntimeError("postgres://user:secret-canary@internal")` handler, then walked the result with `shared.billing_errors.iter_exception_chain`. Result: `['RunningTaskAuthorityIndeterminate|task authorization revalidation is indeterminate']` — chain length 1, no PG text. The mechanism is real: all three sites (`:3384`, `:3457`, `:3495`) assign to `running_authority_error` inside the handler and `raise ... from None` after the `try/except`, so `sys.exc_info()` is already cleared and `__context__` is never populated. The `P0_GRAY_DISABLED` branch returns `detach_authz_error(exc)` and is caught by the module-level `except Exception`, which refunds via `_refund_video_model_call` — no raise, no leak.

**N3 — FIXED.** `FeatureCreditSettlementConflict` is caught explicitly in all three helpers (`:346`, `:379`, `:421`), and the raw `"...: %s", exc` interpolation is gone from both `_confirm_...` and `_refund_...`, replaced with `safe_error_type`. What the new logs still miss is N9.

**N4 — FIXED.** `TaskAuthorityFault` (`envelope.py:83`) is a separate type, and `"fault"` was removed from `TaskAuthorityUnavailable._CODE_BY_FAILURE_KIND` so it can no longer be constructed. `consumer.py:228` raises the fault type and `ports/local/tasks.py:310` keeps `retryable_error_types=(TaskAuthorityUnavailable,)`, so faults are not retried. Semantically consistent with EE: `admission_repository.py:93` classifies transient SQLSTATEs as `Unavailable` and everything else as `Fault`. (I have a separate concern about what CE feeds into that bucket — filed on the EE PR, since the fail-fast lands there.)

**N5 — FIXED.** `video_generator.py:3624` calls `mark_model_call_credit_settlement_for_review(reservation_id, ...)` with provider evidence; the method exists on `NoOpUsageMeter` (`ports/local/usage.py:143`), on the `UsageMeter` protocol, and on EE's `EEUsageMeter`. `reservation_id` / `provider_request_id` / `task_id` are all initialised at `:2943-2945`, so the handler cannot `NameError`. Residual gap is N11.

**N6 — FIXED.** `deadline`, `check_cancelled`, `monotonic` and the `time` import are gone from `authz_retry.py`; no caller in `src/` or `tests/` passes any of them.

**CE/EE seam** matches at SuperTale `8ac6f5fe`: `resolve_feature_credit_reservation` and `mark_model_call_credit_settlement_for_review` are both implemented, the four outcome literals and the "non-resolved forbids reservation_id" invariant match `ports/usage.py:23-38`, `_project_key` argument order matches, and `TaskAuthorityFault` is handled in EE's celery task. Task-row metadata is merged (`task_state.py:281`, EE `task_state_store.py:737`), so stripping the handles from `run_metadata` does **not** wipe the enqueue-written keys the EE terminalizer and cancel path read.

### New findings

**N7. `src/novelvideo/task_backend/run_core.py:686` — high (blocking) — the strip also hits the LLM billing context, disabling EE's paid-execution evidence for every task-run model call, which then makes real spend auto-refundable.**

The stripped `billing_metadata` is passed to `_set_project_task_metrics_context` at `:783`, which calls `set_llm_usage_context(billing_metadata=...)` → `_BILLING_METADATA_CTX`. EE's `metrics_emit.reserve_current_model_call_credit` reads exactly those two keys out of `get_billing_metadata_context()` (`metrics_emit.py:390`) to decide whether to call `begin_paid_execution_attempt`, and `mark_current_paid_execution_attempt` falls back to the same keys (`:596`). EE's producer writes them at `task_backend.py:546-547`, and at merge-base `7ca5efa8:589` they reached the context unstripped.

I confirmed from the EE side that there is no compensating path: with `feature_reservation_id == ""`, `begin_paid_execution_attempt` is never called, `_PAID_EXECUTION_ATTEMPT_CTX` stays empty, and `mark_current_paid_execution_attempt` returns early.

The damage goes past lost telemetry. EE's `resolve_feature_credit_settlement_reviews` feeds `feature_reservation_cost_evidence` into `_review_refund_is_safe` (`credit_settlement.py:65`), which auto-refunds when `attempts == 0`. With no attempt rows ever written, **a task that did burn provider spend now looks like a zero-cost task and is automatically refunded.** The fail-closed `except Exception: raise` guard in `reserve_current_model_call_credit` — which exists specifically to abort the provider call when evidence cannot be persisted — also becomes dead code for task runs.

Suggested fix: after `_resolve_feature_reservation_id` succeeds, re-inject the *resolved trusted* reservation id into the billing context, rather than removing the key outright. `runners/episode_assets.py:73` has the same problem via the stripped envelope.

**N8. `src/novelvideo/task_backend/run_core.py:708` — medium (blocking) — the new resolution call is an unguarded network dependency in the prologue; anything but `FeatureSettlementResolutionRejected` strands the task as `queued` forever.**

Only `FeatureSettlementResolutionRejected` is caught. EE's `EEUsageMeter.resolve_feature_credit_reservation` → `PostgresTaskStateStore.resolve_feature_credit_reservation` performs a synchronous PG call that raises on connection loss, raises `RuntimeError("invalid feature settlement resolution")` on an unexpected `outcome`, and `FeatureSettlementResolution.__post_init__` raises `ValueError` on a malformed row. All of those escape `run_project_task_core_sync` **before** `manager.begin_task_execution_for_project`, so the task-center row never leaves `queued` and `fail_task_for_project` is never called. On the inline lane `_on_background_task_done` (`ports/local/tasks.py:355`) only logs, so the row sits at `queued` indefinitely with its dedup key held — the user cannot re-submit the same task. A five-second PG blip is enough. Wrap the `asyncio.run` in a broad handler that fails the task, leaving the reservation to EE's awaiting-review row, the way every other settlement call in this file is wrapped.

**N9. `src/novelvideo/task_backend/run_core.py:346` (also `:379`, `:421`, `:437`) — low — money-path failure logs no longer identify which reservation failed.**

The new handlers emit `settlement_action`, `safe_error_type` and a freshly minted `error_id` that is never persisted, returned, or attached to the settlement record, while `reservation_id` — an internal ledger id, not sensitive — is absent from `extra`, as is `exc_info`. When `settle_feature_credit_reservation(action="confirm")` fails, the credit stays in `feature_reserved` and the operator's only artifact is `feature_credit_settlement_adapter_failure settlement_action=confirm safe_error_type=OSError error_id=<random hex>`, with no way to find the affected credit row or the cause. Adding `reservation_id` to `extra` restores correlation without reintroducing the adapter text the safe-logging test guards.

**N10. `src/novelvideo/task_backend/run_core.py:717` — low — the `FeatureSettlementResolutionRejected` path skips task metrics.**

Every other failure exit (`:938`, `:967`) calls `await _emit_project_task_metrics(..., outcome="failed")`; the new early return at `:728` does not. So `ambiguous` and `conflict` resolutions — precisely the cases worth alerting on — are invisible in task-outcome metrics, and a systemic key-drift regression between CE's identity and EE's `task_key` would show up as tasks quietly vanishing rather than as a failure-rate spike.

Related, and worth handling together: EE's resolver returns `not_applicable` both for "this task has no reservation" and for "a reservation exists but one of the six ledger join predicates didn't match" (SuperTale `0078:91`). CE turns that into `return ""` at `:252` with no log at all, so the second case is silent on both sides. I've filed the EE half separately; on the CE side a log line when the resolver returns `not_applicable` would make it diagnosable.

**N11. `src/novelvideo/generators/video_generator.py:3639` — low — the N5 review enqueue is best-effort with no fallback, so the case it was added for can still orphan the reservation.**

If `mark_model_call_credit_settlement_for_review` raises — the EE outbox write failing, i.e. the same PG unavailability that likely caused the indeterminate authority in the first place — the exception is swallowed into a log line and `RunningTaskAuthorityIndeterminate` propagates unchanged. The model-call reservation from `:3307` is then neither refunded, confirmed, nor durably parked, and `run_core`, which handles only the *feature* reservation, gets no signal that the model settlement is unresolved. Authority-indeterminate and outbox-write-failure are strongly correlated, so this is not a remote branch.

**N12. `src/novelvideo/authz_retry.py:63` — low — the default `retryable_error_types=(AuthzServiceUnavailable,)` still matches `AuthzServiceFault`, contradicting the split this PR just introduced.**

`AuthzServiceFault` subclasses `AuthzServiceUnavailable` (`ports/authz.py:102`), so `egress_binding.py:67` and `video_generator._revalidate_organization` (`:2783`) keep retrying an error EE classifies as deterministic and non-transient (`admission_repository.py:93`), while `ports/local/tasks.py:310` now correctly treats the identical failure as terminal. A malformed admission row therefore makes the post-accept revalidation sleep through the full jitter budget on a task that was never going to be admitted, delaying the terminal transition of an already-accepted provider operation. Passing `retryable_error_types=(AuthzServiceUnavailable,)` with an explicit `AuthzServiceFault` exclusion — or reworking the hierarchy so `Fault` is not an `Unavailable` — would make the two boundaries agree.

Handanhhhy:
Addressed the latest review findings N7/N8 in 93e23dc0. The runner now rebuilds billing context only from the authoritative reservation resolver, and unexpected resolver failures fail the exact task closed before runner execution without making a settlement decision. Added adversarial coverage for forged handles and resolver outages. Verification: 106 passed, 1 deselected; Black, Ruff, and diff checks passed. Please re-review.

lywaterman (commented):
## Follow-up review 3 — `93e23dc0`

Re-reviewed at head `93e23dc0`, alongside EE at SuperTale#66 head `e1e67e39`. Focused run over `test_running_task_authz_indeterminate.py`, `test_trusted_task_settlement_resolution.py`, `ports/test_tasks.py`, `test_task_envelope_consumer.py`, `test_authz_retry.py`, `test_feature_settlement_safe_logging.py`, `test_task_timeout.py`, `test_p0g4c_video_egress.py`: **107 passed, 0 deselected**, 4.83 s. I did not reproduce the "1 deselected".

**First, a correction to my previous review.** I wrote that N7's consequence chain ended in EE auto-refunding whenever `attempts == 0`, i.e. that the hourly settlement CronJob would systematically refund real spend. That was too strong and I withdraw it. `_review_refund_is_safe` (`credit_settlement.py:52`) refunds at `attempts == 0` only when there is also a *positive* durable marker — `cancel_requested is True`, or `source` in `_SAFE_INTERRUPTION_SOURCES`. It is evidence of a business-level abort, not absence of evidence. N7 was a real bug (missing paid-execution evidence, and a fail-closed guard turned into dead code), but it did not produce automatic wrong refunds. Sorry for the overstatement.

### Verdicts

**N7 — FIXED.** Verified by runtime trace rather than by the tests. `_trusted_metrics_billing_metadata` (`run_core.py:233`) strips both delivery handles and then re-injects the id returned by `_resolve_feature_reservation_id`; the result is assigned to `envelope["billing_metadata"]` (`:775`) and passed to `_set_project_task_metrics_context` (`:830`). Driving the real, un-monkeypatched `run_project_task_core_sync` with a delivery carrying `feature_credit_reservation_id="forged"` / `feature_credit_charge_id="forged2"` and a resolver returning `trusted-res-9`:

```
envelope_bm:    {'feature_key':…, 'model_call_credit_policy':…, 'feature_credit_reservation_id': 'trusted-res-9'}
ctx_in_runner:  {…, 'feature_credit_reservation_id': 'trusted-res-9'}   # _BILLING_METADATA_CTX
ctx_in_coro:    {…, 'feature_credit_reservation_id': 'trusted-res-9'}   # survives asyncio.run
settle:         ('trusted-res-9', action='confirm')
```

So the value reaching EE's `get_billing_metadata_context()` is the resolved trusted one, visible both to sync runner code and inside `asyncio.run` sub-contexts. On `charge_id`: it is not restored, but nothing short-circuits — `metrics_emit.py:390-392` and `:596-598` both read `reservation_id or charge_id`, and EE's producer sets both to the identical value (`task_backend.py:546-547`), so the `or` resolves on the first key. `begin_paid_execution_attempt` fires, `_PAID_EXECUTION_ATTEMPT_CTX` is populated, and `feature_reservation_cost_evidence` returns `attempts > 0`. `runners/episode_assets.py:60/73` reads the same rehydrated dict, so that half is fixed too.

**N8 — FIXED, with one residual.** The new `except Exception` at `:740` calls `manager.fail_task_for_project`, which sets `status = "failed"` (`task_state.py:1243`); since `ACTIVE_PROJECT_TASK_STATUSES = {"submitting","queued","running"}` (`task_state.py:44`), the dedup key *is* released. Both lanes are covered because `run_project_task_core_sync` is the single shared entrypoint (`ports/local/tasks.py:335`, EE `celery_tasks.py:390`), and EE's celery lane consumes the returned dict without retrying. The new unguarded `asyncio.run(_emit_project_task_metrics(...))` is safe — that function swallows everything into a `logger.debug` (`:538`). Residuals are N14 and N15.

**N9–N12 — untouched, as expected.** `93e23dc0` changes only `run_core.py` plus two test files. I re-read `:359-450` (N9 — still `settlement_action` + `safe_error_type` + a throwaway `error_id`, no `reservation_id`, no `exc_info`) and `:722-739` (N10 — the `FeatureSettlementResolutionRejected` branch still returns without `_emit_project_task_metrics`, which its new sibling branch *does* call, so the asymmetry is now visible within one handler chain).

**The new tests are not vacuous.** `_delivery()` supplies `billing_metadata={"feature_credit_reservation_id": "foreign-reservation"}`, so the forged-handle assertions have something to catch; reverting the fix breaks them two different ways (equality mismatch, or `KeyError` when the stripped dict is falsy and the key is never set). The runner-side `assert` propagates into the generic failure path and changes `error_code`, so a silent pass is not possible. `test_settlement_resolution_fault_fails_task_before_runner` genuinely asserts `invoked is False` and `metric_outcomes == ["failed"]`.

### New findings

**N13. `run_core.py:233` — medium — `_trusted_metrics_billing_metadata` sanitizes 2 keys and republishes the rest as trusted.**

The docstring says "rehydrate private billing context from authoritative settlement state", but only `feature_credit_reservation_id` is authoritative. Everything else in the dict — `model_call_credit_policy`, `feature_key`, `feature_credit_cost` — is copied verbatim from `delivery.billing_metadata`, which is still outside `_SIGNED_PAYLOAD_FIELDS` (`consumer.py:36`), and is now *deliberately* pushed into `_BILLING_METADATA_CTX`.

EE reads `model_call_credit_policy` at `metrics_emit.py:387`, and at `:418` does `if policy == MODEL_CALL_CREDIT_POLICY_FEATURE_INCLUDED: return ""` — skipping `reserve_model_call_credits` entirely. Today every `FeatureBillingSpec` uses the default `feature_included` (no spec in `feature_billing.py` overrides it), so this is latent rather than live. But the first spec that opts into per-call charging turns a forged `model_call_credit_policy: "feature_included"` into free model calls, via exactly the envelope-replay vector this PR just closed for reservation ids. `feature_key` similarly lands in `begin_paid_execution_attempt(feature_key=…)` (`metrics_emit.py:404`), so an attempt-evidence row can be labelled with a feature unrelated to the resolved reservation.

Suggest an explicit allowlist of observational keys, and deriving the policy from the resolved reservation's feature spec rather than from the envelope.

**N14. `run_core.py:740` — medium (blocking) — failing closed without a settlement decision leaves the user's credits debited indefinitely.**

The handler fails the task and returns without touching the reservation — it has no id, which is the whole point of the branch. But the reservation's outbox row was created at reserve time with `metadata = {"source": "feature_reserve", …}` and no `business_outcome` (`credit_ledger.py:2475-2490`). `resolve_feature_credit_settlement_reviews` (`credit_settlement.py:541-552`) refunds only when `business_outcome ∈ {cancelled, failed, not_delivered}` or `source ∈ _SAFE_INTERRUPTION_SOURCES = {sync_api, sync_api_interrupted, task_cancel_api, task_cancelled, task_failed}`. `feature_reserve` is in neither, so every hourly pass calls `_defer_feature_credit_settlement_review` and increments `awaiting` forever.

Concretely: a five-second Postgres blip during resolution now leaves the user debited for a task that never started, with no automatic recovery; the user resubmits and is charged a second time. This is strictly better than the pre-fix behaviour, which stranded the row *and* the credits — but the money half of N8 is not closed. CE cannot fix it alone; it needs an EE rule keyed off the terminal task row / `error_code`, or a CE call that parks the reservation by verified identity rather than by handle.

**N15. `run_core.py:749` — low — the fail-closed write shares its backing store with the call that just failed, so the motivating scenario can still strand the row at `queued`.**

On the EE lane `manager.fail_task_for_project` is a Postgres write. The scenario N8 was filed for is a transient PG outage: `resolve_feature_credit_reservation` raises `ConnectionError`, control reaches `:749`, and the terminalization write fails with the same error — which is *not* caught, so it escapes `run_project_task_core_sync` and the row sits at `queued` with its dedup key held, exactly as before. The new test cannot detect this because its `Manager.fail_task_for_project` is an in-memory `list.append`. A bounded retry around the terminalization, or a `try/except` that logs and still returns the failure dict, would make the guarantee match the claim. (This is a property of the other failure paths too, but they at least reach `running` after `begin_task_execution_for_project` has already proved the store is reachable.)

Handanhhhy:
Addressed the latest follow-up findings N13–N15 and CI at head 88ad5e6c.\n\n- Billing context is built exclusively from an identity-bound authoritative snapshot.\n- Resolver failures remain runner-free and reuse the signed task.settlement terminalization lane.\n- Broker dispatch failures use the existing exponential retry seam.\n- Added the two new settlement test files to the generated license inventory.\n- DCO sign-off is present on every commit in the PR range.\n\nVerification:\n- Targeted settlement suite: 108 passed, 1 environment-only deselection.\n- License inventory regression: passed.\n- Previously timing-sensitive concurrency test: passed 5 consecutive local runs and both full CI runs.\n- GitHub CI: both deps jobs passed; DCO, lint, gitleaks, dependency review, and policy checks passed.\n\nPlease re-review together with claymorelab/SuperTale#66.

lywaterman (commented):
## Follow-up review 4 — `88ad5e6c`

Re-reviewed at head `88ad5e6c`, alongside EE at SuperTale#66 head `a55584b3`. The branch was force-pushed with DCO sign-offs, so SHAs changed; the old head object is still available locally, and `git diff 93e23dc0..HEAD` (5 files, +157/-72) is the real tree delta for this round.

Focused run over `test_running_task_authz_indeterminate.py`, `test_trusted_task_settlement_resolution.py`, `ports/test_tasks.py`, `test_task_envelope_consumer.py`, `test_authz_retry.py`, `test_feature_settlement_safe_logging.py`, `test_task_timeout.py`, `test_p0g4c_video_egress.py`: **109 passed, 0 deselected**, 3.84 s. I did not reproduce the claimed "108 passed, 1 deselected".

> **Blocking:** F1 and F3. Both charge the user money that shouldn't be charged, and both are new in this round.

### Verdicts

**N13 — FIXED.** The rehydration helper is gone; the billing context is now built solely by `FeatureSettlementResolution.trusted_billing_metadata()` (`ports/usage.py:60`). Exactly three keys reach `_BILLING_METADATA_CTX`, and I traced each origin: `feature_credit_reservation_id` and `feature_key` come from `ee_resolve_feature_settlement` (SuperTale `0078`), whose candidate query joins `credit_settlement_outbox` to `credit_ledger` pinned on `ledger.user_id = p_requester_user_id`, `ledger.project_id`, `ledger.metadata->>'task_id' = p_expected_task_id` and `ledger.metadata->>'task_type' = states.task_type`, with `feature_key := ledger.model` — the value EE itself wrote at reserve time (`credit_ledger.py:2459`; `model=feature_key` on the org path at `:2377`).

`model_call_credit_policy` no longer travels at all: `EEUsageMeter.resolve_feature_credit_reservation` looks the resolved `feature_key` up in the compile-time `FEATURE_BILLING_SPECS` registry (`ports_bootstrap.py:206`) and raises if there is no spec. Neither field is envelope-derived any more, and the identity feeding the resolver is entirely HMAC-covered. `feature_credit_cost` and `feature_credit_charge_id` are dropped safely: no EE reader for the former, and `metrics_emit.py:390/596` read `reservation_id or charge_id`, so the `or` resolves on the first key. The residual is the *opposite* direction — F1 and F7.

**N14 — FIXED, for the resolver-failure lane, via EE.** Traced end to end: CE returns `{"failed": True, "error_code": "FEATURE_SETTLEMENT_RESOLUTION_FAILED"}` → EE `celery_tasks.py:409` signs a terminalization marker and sends it to the `task.settlement` queue → `terminalize_rejected_delivery` → `ee_terminalize_undelivered_task`. Because CE already wrote `status='failed'`, the function takes the `v_status = 'failed' AND v_existing_error IS NOT DISTINCT FROM p_error` replay branch (`0078:234`), falls through to the resolver, stamps `business_outcome='not_delivered'` on the outbox row, and returns `terminalized_evidence_ready` plus the reservation id; celery then calls `refund_undelivered_feature_credit_reservation`. The money **is** refunded.

Predicate check against SuperTale `a55584b3`: `_SAFE_INTERRUPTION_SOURCES` is unchanged and still lacks `feature_reserve`, but that no longer matters — `_REFUNDABLE_BUSINESS_OUTCOMES` contains `not_delivered` (`credit_settlement.py:34`), so the hourly review resolver is a working backstop if the direct refund fails. Two caveats: F2 and F3.

**N15 — PARTIALLY FIXED / relocated.** The `try/except` around `manager.fail_task_for_project` stops the escape, but it does not make the row terminal. The actual durability now comes from EE's post-return dispatch: with the fail write dead, the row is still `queued`, and the terminalizer's `IF v_status IN ('submitting','queued')` branch (`0078:214`) terminalizes and refunds it. That path lives entirely in EE and reaches CE only through a return-value convention — see F4.

**N9–N12 — untouched, as expected.** The delta touches 5 files (`license-inventory.csv`, `ports/usage.py`, `run_core.py`, 2 tests). `run_core.py:346/379/421` still log `settlement_action` + `safe_error_type` + a throwaway `error_id`, with no `reservation_id` and no `exc_info` (N9); the `FeatureSettlementResolutionRejected` branch at `:722` still returns without `_emit_project_task_metrics` (N10); `video_generator.py` and `authz_retry.py` are not in the diff (N11, N12).

### New findings

**F1. `run_core.py:763` — high (blocking) — dropping the envelope-derived `model_call_credit_policy` with no fallback turns every non-`resolved` outcome into per-model-call charging on top of the feature charge.**

`trusted_billing_metadata()` returns `{}` unless `outcome == "resolved"`, so the billing context loses `model_call_credit_policy` entirely. EE's `reserve_current_model_call_credit` short-circuits only on `if model_call_credit_policy == MODEL_CALL_CREDIT_POLICY_FEATURE_INCLUDED: return ""` (`metrics_emit.py:418`); with the key absent it falls through to `reserve_model_call_credits`. At merge-base and at the previous head, the producer-set `feature_included` (EE `task_backend.py:509`) reached the context and suppressed that.

Confirmed by driving the real `run_project_task_core_sync` with a resolver returning `not_applicable`:

```
llm_ctx == {'billing_user_id': …, 'requester_user_id': …, 'billing_task_type': …}
# no model_call_credit_policy, no feature_key
```

Reachable today: (a) a free-priced feature — `allow_free=True` specs pass the `unit != "free"` guard at `credit_ledger.py:1191`, `cost <= 0` returns `{"id": "", "reserved": False}` at `:2320`, no outbox row is written, and the resolver says `not_applicable`; (b) any redelivery where the outbox row is no longer `status='awaiting'` (e.g. after the reaper already settled it), or any drift in the six join predicates — my earlier review established that `not_applicable` doubles as "a reservation exists but a predicate didn't match". In those cases the user is debited the feature credit *and* the per-call credits.

**F2. `run_core.py:739` — medium — the refund after a resolver failure hangs on a bare cross-repo string literal that nothing pins.**

Because CE terminalizes first, EE's `ee_terminalize_undelivered_task` can only reach the refund through the replay branch, which requires `states.error IS NOT DISTINCT FROM p_error`, where `p_error` is EE's `_ERROR_BY_REASON["settlement_resolution_failed"][1]`. CE writes `error="feature settlement resolution failed"` as an inline literal; EE hard-codes the same string in `task_terminalization.py:48`. There is no shared constant and no CE test asserting the value — only EE's `test_task_terminalization_marker.py:61`.

If the message is ever reworded, prefixed, or localized, the function returns `task_already_failed_other` and returns *before* the outbox update: the row keeps `source: "feature_reserve"` with no `business_outcome`, `_review_has_refundable_business_outcome` stays false, and the reservation is deferred hourly forever. That is exactly the N14 failure, restored silently by an edit that looks cosmetic.

**F3. `run_core.py:722` — medium (blocking) — the `ambiguous` / `conflict` sibling branch still fails closed with no settlement decision and gets no terminalization.**

EE dispatches the terminalization marker only when `result.get("error_code") == "FEATURE_SETTLEMENT_RESOLUTION_FAILED"` (`celery_tasks.py:409`); this branch returns `FEATURE_SETTLEMENT_RESOLUTION_AMBIGUOUS` / `_CONFLICT`. On `ambiguous` there are two or more awaiting feature reservations tied to this exact `task_id` (the resolver's `v_count > 1` path), the task never starts, and neither row gets a `business_outcome` — `feature_reserve` is not in `_SAFE_INTERRUPTION_SOURCES`, so `resolve_feature_credit_settlement_reviews` defers both every hour indefinitely. The user is debited **twice** for a task that never ran, and a resubmit charges again. This is N14's exact shape surviving in the branch that was not routed.

**F4. `run_core.py:750` — medium — the swallowed terminalization failure moves the single point of failure into EE and leaves CE's own lane quieter than before.**

The `except Exception` logs and continues, but the row is still `queued` with its dedup key held. Recovery exists only because EE inspects the returned dict afterwards. CE's own caller discards it: `ports/local/tasks.py:332` does `await loop.run_in_executor(..., run_project_task_core_sync, ...)` and `_on_background_task_done` only logs. So on the inline lane the task is now stranded *and* silent, where before it at least surfaced an exception. The new contract — "the caller must dispatch terminalization when `error_code == FEATURE_SETTLEMENT_RESOLUTION_FAILED`" — is undocumented in CE, unenforced, and not covered by any CE test. At minimum, document it on `run_project_task_core_sync` and add a bounded retry around the fail write, so the guarantee does not depend on an out-of-repo caller.

**F5. `tests/test_running_task_authz_indeterminate.py:239` — low — the new test asserts none of what its name claims.**

`test_resolution_failure_remains_durably_owned_when_fast_path_write_fails` only checks the return dict and `metric_outcomes == ["failed"]`. Nothing in it observes ownership, terminalization, or the task row — after `FailingManager.fail_task_for_project` raises, the row is by construction *not* durably owned in this test's world. It also plants two DSNs with secrets (`secret-canary`, `second-secret`) and, unlike its sibling `test_settlement_resolution_fault_fails_task_before_runner`, never asserts they are absent from the logs or the failure records, so the leak surface the canaries exist for is unchecked. Rename it to what it tests — that the fast-path write failure is swallowed — and add the canary assertion.

**F6. `run_core.py:322` — low — `feature_credit_reservation_id()` is now dead code that reads the exact untrusted handle this PR closed off.**

After this round no caller remains in `src/`, `tests/`, or SuperTale's `src/`. It is a module-level, public-looking helper whose whole job is to pull `feature_credit_reservation_id` / `feature_credit_charge_id` out of enqueue-side delivery metadata — the envelope-replay vector the resolver replaces. Leaving it exported invites reintroducing the bug in the next runner. Delete it, or rename it so the untrusted provenance is unmissable.

**F7. `ports/usage.py:60` — low — a partial `resolved` snapshot fails open toward charging instead of closed.**

`trusted_billing_metadata()` filters out empty values, and `__post_init__` requires only `reservation_id` when `outcome == "resolved"`. So a resolution carrying a reservation id but an empty `model_call_credit_policy` yields a context that passes EE's `feature_reservation_id` check — attempt evidence is written — and then fails the `feature_included` comparison at `metrics_emit.py:418`, charging per-call credits on top of the feature reservation. That is precisely the shape of a CE-ahead-of-EE deploy window, since EE only started populating the field in `94626bd9`. Given the deploy-order ban is documented but not machine-enforced, it would be cheap to make `resolved` require both `feature_key` and `model_call_credit_policy` — or default the policy to `feature_included` — so a partial snapshot fails the task rather than billing the user.

Handanhhhy:
Superseded by a two-way split. This PR stays in draft as the historical record
of the review rounds; please review the split PRs instead.

**Lane A — the authz transient retry closure, what this PR's title describes:**
- CE: #324 ← merge first
- EE: claymorelab/SuperTale#76

Verified green: 1 failed / 3329 passed, where the single red is identical on
`origin/staging` and passes in isolation on both refs.

**Lane S — the paid-execution settlement convergence, which grew in during
review:** not yet opened. It carries every blocking finding from rounds 1–5
(F1, F3, and EE's N14), and none of them are reachable from Lane A — Lane A
does not touch `run_core.py`, `ports/usage.py`, or any settlement path.

The split exists because five review rounds in ~25 hours did not converge: each
round shipped its fixes as new code, which the next round then reviewed as new
code. Separating the two lanes lets the retry closure land on its own evidence
instead of waiting on the settlement work.

Handanhhhy:
Superseded by the split Lane A and Lane S PRs, both now merged: #324 and #326. Closing this original combined draft without merging.

PR #316, opened 2026-08-19
fix(egress): avoid organization context for platform tasks

wangwenqq:
## PR 类型 / Type of PR
- [x] 🐞 Bug 修复 / Bug fix

## 改动说明 / What this PR does and why
普通/平台任务在 Freezone 网络叶子分发时也被注入了 `egress_context`，音乐节点会将该非组织身份判为非法组织出网身份，导致任务报 `ORG_EGRESS_DENIED`。

本 PR 仅在组织任务中注入受信任的组织出网上下文；平台和个人任务保持无组织上下文的既有直连路径。同时补齐回归测试、既有测试桩和许可证清单。

## 关联 issue / Linked issues
N/A

## 给 reviewer 的说明 / Notes for the reviewer
组织任务仍携带上下文，继续走网关、组织凭证与组织账本；普通任务不再误触发组织闸门。

## 面向用户的变更 / User-facing change
```release-note
修复普通用户生成音乐时错误提示 ORG_EGRESS_DENIED 的问题。
```

## 自检 / Checklist
- [x] 已自测，相关 pytest 通过 / Self-tested
- [x] 必要的文档已更新 / Docs updated where needed
- [x] 提交信息清晰（建议 Conventional Commits）/ Clear commits
- [x] 每个 commit 都已 `git commit -s` 附 DCO 签署 / Every commit signed off with DCO
- [x] 我已阅读并同意贡献者协议 / I agree to contributor terms

### DIFF
--- a/src/novelvideo/task_backend/runners/freezone.py
+++ b/src/novelvideo/task_backend/runners/freezone.py
@@ -156,7 +156,11 @@ def _call_freezone_leaf(
         if rule.egress is LeafEgress.LOCAL:
             return leaf(**kwargs)
         if rule.egress is LeafEgress.NETWORK:
-            return leaf(egress_context=context, **kwargs)
+            # 组织任务必须携带受信任的组织身份以走网关；平台与个人任务
+            # 保持无组织上下文的既有直连语义，不能误触发组织出网闸门。
+            if context is not None and context.is_organization:
+                return leaf(egress_context=context, **kwargs)
+            return leaf(**kwargs)
     # 未分类与 DENIED 同待遇：默认 fail-closed。
     if context is not None and context.is_organization:
         raise InvalidTaskEnvelope() from None
--- a/tests/test_freezone_leaf_egress.py
+++ b/tests/test_freezone_leaf_egress.py
@@ -1,0 +1,33 @@
+from types import SimpleNamespace
+
+from novelvideo.task_backend.runners import freezone
+
+
+def test_network_leaf_uses_organization_context_only(monkeypatch):
+    platform_context = SimpleNamespace(is_organization=False)
+    monkeypatch.setattr(
+        freezone,
+        "_extract_trusted_egress_context",
+        lambda _envelope: platform_context,
+    )
+
+    def leaf(**kwargs):
+        return kwargs
+
+    assert freezone._call_freezone_leaf({}, leaf, "generate_freezone_audio_eleven_music") == {}
+
+
+def test_network_leaf_passes_organization_context(monkeypatch):
+    organization_context = SimpleNamespace(is_organization=True)
+    monkeypatch.setattr(
+        freezone,
+        "_extract_trusted_egress_context",
+        lambda _envelope: organization_context,
+    )
+
+    def leaf(**kwargs):
+        return kwargs
+
+    assert freezone._call_freezone_leaf(
+        {}, leaf, "generate_freezone_audio_eleven_music"
+    ) == {"egress_context": organization_context}
--- a/license-inventory.csv
+++ b/license-inventory.csv
@@ -1869,6 +1869,7 @@ tests/test_freezone_canvas_store.py,Elastic-2.0,2026 SuperTale contributors,REUS
 tests/test_freezone_generation_history_prompt.py,Elastic-2.0,2026 SuperTale contributors,REUSE.toml aggregate annotation
 tests/test_freezone_image_backend.py,Elastic-2.0,2026 SuperTale contributors,REUSE.toml aggregate annotation
 tests/test_freezone_image_node_backend.py,Elastic-2.0,2026 SuperTale contributors,REUSE.toml aggregate annotation
+tests/test_freezone_leaf_egress.py,Elastic-2.0,2026 SuperTale contributors,REUSE.toml aggregate annotation
 tests/test_freezone_mainline_enqueue_projection.py,Elastic-2.0,2026 SuperTale contributors,REUSE.toml aggregate annotation
 tests/test_freezone_mark_backend.py,Elastic-2.0,2026 SuperTale contributors,REUSE.toml aggregate annotation
 tests/test_freezone_mark_detect_egress.py,Elastic-2.0,2026 SuperTale contributors,REUSE.toml aggregate annotation
--- a/tests/test_freezone_image_backend.py
+++ b/tests/test_freezone_image_backend.py
@@ -3502,17 +3502,14 @@ async def test_freezone_celery_text_runner_records_project_node_history(
     ctx = _project_ctx(tmp_path)
     project_dir = ctx.output_dir
 
-    # 真 leaf 出网，签名里有 egress_context（`freezone/text_node.py`）。个人信封下
-    # 它是 None，但仍会传进来——分发按分类注入，不再看替身的签名长什么样。
+    # 普通任务不传组织身份；组织任务才由分发层注入 egress_context。
     async def fake_translate_freezone_text(
         *,
         text: str,
         node_type: str,
-        egress_context,
     ):
         assert text == "你好"
         assert node_type == "text"
-        assert egress_context is None
         return "hello", "zh", "en"
 
     class FakeTaskManager:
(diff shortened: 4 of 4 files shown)


src/novelvideo/task_backend/runners/freezone.py
```diff
@@ -156,7 +156,11 @@ def _call_freezone_leaf(
         if rule.egress is LeafEgress.LOCAL:
             return leaf(**kwargs)
         if rule.egress is LeafEgress.NETWORK:
-            return leaf(egress_context=context, **kwargs)
+            # 组织任务必须携带受信任的组织身份以走网关；平台与个人任务
```
lywaterman, line 159: **medium (1) — the root cause is `freezone/audio_node.py:557`, and it is still wrong after this change.**

```python
# audio_node.py:557
if egress_context is None:
    egress_context = ambient_organization_egress_context()
if egress_context is None:
    generator = IndexTTS2FalClient()
else:
    generator = IndexTTS2FalClient(provider="newapi", egress_context=egress_context)
```

That branches on `is not None`, while the other leaves branch on `is_organization` — `freezone/jobs.py:78-83` is the correct shape:

```python
if egress_context is not None and egress_context.is_organization and provider == "volcengine":
    raise EgressError("ORG_EGRESS_DENIED")
```

This PR suppresses the symptom at the dispatch layer, and that works today only because `_call_freezone_leaf` is the sole caller that ever handed a platform context to these leaves. The leaf itself remains a trap: any future caller passing a platform context directly gets `ORG_EGRESS_DENIED` again.

Adding `and egress_context.is_organization` at `audio_node.py:557` would make the leaf correct on its own — and would let the dispatch keep injecting identity explicitly instead of dropping it.

---

**medium (2) — this erases a distinction the design keeps on purpose.**

`egress_context.py:68-72` states the rationale directly: collapsing 「平台任务，允许」 and 「调用点忘了穿参数」 into the same `None` is what leaves the gate unable to tell them apart, and the ambient fallback exists to compensate.

`_call_freezone_leaf` was the one place guaranteeing a NETWORK leaf always received an explicit value. After this change a platform task reaches the leaf with the argument **absent**, so correctness now depends on `run_core.py`'s ambient binding being present and correctly scoped for every platform delivery. That is a real dependency shift, and nothing in this PR tests it.

Fixing at the leaf (above) avoids the shift entirely; if you prefer to keep the dispatch-layer fix, a test asserting the ambient binding stays non-organization for a platform delivery would pin the new dependency.

---

**low (3) — the comment describes the prior behavior inaccurately.**

> 平台与个人任务保持无组织上下文的**既有**直连语义

For platform tasks the *existing* behavior was not a direct connection — it was `ORG_EGRESS_DENIED` from `indextts2_fal.py:189`. This restores the **intended** semantics rather than preserving the existing ones. In the file that is itself the egress gate, that distinction is worth stating precisely.

tests/test_freezone_leaf_egress.py
```diff
@@ -0,0 +1,33 @@
+from types import SimpleNamespace
+
+from novelvideo.task_backend.runners import freezone
+
+
+def test_network_leaf_uses_organization_context_only(monkeypatch):
+    platform_context = SimpleNamespace(is_organization=False)
+    monkeypatch.setattr(
```
lywaterman, line 8: **low — this test asserts the plumbing, not the failure the PR fixes.**

Two gaps:

1. `monkeypatch.setattr(freezone, "_extract_trusted_egress_context", lambda _envelope: platform_context)` with a `SimpleNamespace` bypasses the real type guard at `runners/freezone.py:48` (`if type(context) is not TrustedEgressContext: raise InvalidTaskEnvelope`). The test would not catch a regression in the real `TrustedEgressContext` — including a change to how `is_organization` derives from `billing_principal.kind`.
2. It stops at "which kwargs did the fake leaf receive". The bug actually manifests two layers down, where `audio_node.py:557` picks the client and `indextts2_fal.py:189` returns `ORG_EGRESS_DENIED`.

A case at the `generate_freezone_audio_speech` level — platform delivery, assert `IndexTTS2FalClient` is constructed **without** `provider="newapi"` — would pin the behavior that actually broke. Using a real `TrustedEgressContext` with a platform `BillingPrincipal` instead of `SimpleNamespace` would also let this test survive a refactor of the context type.

lywaterman (commented):
## Review

**The diagnosis is right and the fix works** — but the root cause is not in the line being changed.

### The bug is real, and platform tasks were hard-failing (not just mis-routed)

I traced the full chain:

1. `_call_freezone_leaf` passed `leaf(egress_context=context, ...)` unconditionally for every NETWORK leaf.
2. A platform task's `TrustedRunnerEnvelope` carries a **non-organization** `TrustedEgressContext` (`is_organization` derives from `billing_principal.kind == "organization"`).
3. `freezone/audio_node.py:557` branches on `if egress_context is None` — anything non-`None` builds `IndexTTS2FalClient(provider="newapi", egress_context=...)`.
4. `generators/indextts2_fal.py:186-189` returns `ORG_EGRESS_DENIED` when `context is not None and not context.is_organization`.

So before this PR, platform Freezone TTS did not "wrongly go through the org gateway" — it **failed outright**. This PR fixes that.

### Verified as safe

- All 12 NETWORK leaves declare `egress_context: TrustedEgressContext | None = None`, so omitting the kwarg cannot raise `TypeError`. I checked each one (`text_node`, `audio_node`, `image_node`, `jobs`, `control_frame_to_sketch`).
- The organization path is unchanged: the context is still injected, and both checks in `indextts2_fal.generate` still pass.
- The direct HTTP caller at `api/routes/freezone.py:6584` never passed the kwarg — it was already on the `ambient_organization_egress_context()` path — so there is no second exposure of the same bug.
- The `LOCAL` branch and the unclassified/`DENIED` fail-closed branch are untouched.

### Findings

| # | Location | Severity |
|---|----------|----------|
| 1 | `freezone/audio_node.py:557` | medium — the actual root cause, still wrong |
| 2 | `runners/freezone.py:159` | medium — erases a distinction the design deliberately kept |
| 3 | `runners/freezone.py:159` (comment) | low — describes the prior behavior inaccurately |
| 4 | `tests/test_freezone_leaf_egress.py` | low — does not cover the failure it fixes |

None of these block merging — the bug is genuinely fixed. Details inline.

src/novelvideo/task_backend/runners/freezone.py
lywaterman: **Correction — withdrawing findings 1 and 2 above. Neither should be changed.**

I went back and audited every branch on `egress_context` across the NETWORK leaves, and I had only looked at part of the picture:

```
audio_node.py:645   non-None and not organization  -> raise RuntimeError("ORG_EGRESS_DENIED")
audio_node.py:823   non-None                       -> adds egress_context, takes the newapi path
jobs.py:538         non-None                       -> raise EgressBoundaryError("ORG_SERVICE_EGRESS_DENIED")
```

So "a platform identity must never reach these leaves" is not an oversight at `audio_node.py:557` — it is a consistent convention across at least four sites in two files, and `ambient_organization_egress_context()` documents it explicitly:

> 多数闸门只分「组织 / 其余」两支，平台身份在那里与 `None` 同义……若回落把平台身份塞进去，平台流量会被自己的组织闸门拒掉——修 fail-open 反倒修出 fail-closed 的误伤。

**Finding 2 was backwards.** I claimed the dispatch change erases a distinction the design deliberately keeps. It does not — on these leaves a platform identity is *defined* to be synonymous with `None`, so this PR is aligning the dispatch layer with the existing rule rather than working around it. And since the ambient fallback only ever returns organization contexts, a platform delivery can only land on `None`. A case where the delivery is platform but ambient is organization would be a bug in the ambient scoping itself, and it would already affect the direct HTTP caller at `api/routes/freezone.py:6584`, which never passed the kwarg either.

**Finding 1 should not be acted on.** Given that convention, `is not None` is a legitimate shorthand for "organization" in these leaves. Adding `and egress_context.is_organization` at `audio_node.py:557` would contradict line 645 three statements later, and doing it properly would mean touching all four sites — a larger change than this PR, for no behavioral gain. The difference from `jobs.py`'s spelling is style, not a defect.

The two low findings still stand and are cheap: the comment describes the prior behavior as 「既有直连语义」 when it was actually `ORG_EGRESS_DENIED`, and the new test stops short of the layer where the bug manifests. Neither blocks merging.

Sorry for the noise — the fix is at the right layer as written.

lywaterman (approved):

PR #320, opened 2026-08-19
feat(models): enforce organization model visibility

ryanhuang-duat:
## 目的

让组织模型可见性同时作用于模型列表和真实任务提交，防止只隐藏前端后仍可绕过调用。

## 改动

- 模型列表按已认证 user_id 获取组织有效目录
- 图片和视频任务在计费、入队前再次强制校验模型可用性
- EE 作用域解析失败时关闭访问，不回退到平台全量目录
- CE 未安装控制面端口时保持原有本地行为
- 不修改积分、预留、扣费、退款或结算代码

## 验证

- 新增 A/B 用户目录隔离与直接提交绕过测试
- 相关测试：21 passed
- 扩展测试：240 passed
- 全量测试：3256 passed，16 skipped，2 deselected；唯一环境项通过固定本地 worker id 后单测 passed
- Ruff 与 diff check：passed

依赖控制面 PR：https://github.com/claymorelab/SuperTale/pull/71

### DIFF
--- a/license-inventory.csv
+++ b/license-inventory.csv
@@ -1545,6 +1545,7 @@ src/novelvideo/ports/local/projection.py,Elastic-2.0,2026 SuperTale contributors
 src/novelvideo/ports/local/release_feed.py,Elastic-2.0,2026 SuperTale contributors,REUSE.toml aggregate annotation
 src/novelvideo/ports/local/tasks.py,Elastic-2.0,2026 SuperTale contributors,REUSE.toml aggregate annotation
 src/novelvideo/ports/local/usage.py,Elastic-2.0,2026 SuperTale contributors,REUSE.toml aggregate annotation
+src/novelvideo/ports/media_model_catalog.py,Elastic-2.0,2026 SuperTale contributors,REUSE.toml aggregate annotation
 src/novelvideo/ports/model_credentials.py,Elastic-2.0,2026 SuperTale contributors,REUSE.toml aggregate annotation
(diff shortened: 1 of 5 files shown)


Handanhhhy:
Review on head `e64ef295059872f18178f0bf5e241a4ff0be77aa` found two issues.

**P1 — CI is currently red.** Both `deps` runs fail in `test_license_inventory_covers_current_git_index` because these new files are missing from `license-inventory.csv`:

```text
src/novelvideo/ports/media_model_catalog.py
tests/ports/test_media_model_catalog_scope.py
```

**P2 — video submission has a model-visibility revocation race.** `_start_or_enqueue_freezone_video_gen` checks visibility at lines 385–390, then awaits media-duration probing at lines 407–422, builds billing at line 455, and enqueues at line 494. If the organization loses access while the async probe is running, the request is still billed/enqueued without another authoritative check.

I added a minimal review-only regression test named `test_video_rechecks_visibility_after_async_media_probe_before_enqueue`:

1. model is initially visible to `user-a`;
2. the first scoped check succeeds;
3. the mocked duration probe disables the model;
4. expected: a 409 before billing/enqueue;
5. actual: `enqueue_project_task()` is called.

Combined focused result:

```text
1 failed, 21 passed
FAILED ...test_video_rechecks_visibility_after_async_media_probe_before_enqueue
AssertionError: task was enqueued after model access was revoked
```

The existing steady-state A/B isolation tests pass. Suggested regression shape: perform the final scoped check after all awaited preprocessing and immediately before billing/enqueue, or enforce it at the admission boundary if atomic revocation semantics are required.

Handanhhhy (approved):

PR #324, opened 2026-08-20
close the authz transient retry path (1/2 of #311)

Handanhhhy:
This is the first half of a two-way split of #311. It carries the change the
original PR title describes — closing the authz transient-retry path — and
nothing else. The paid-execution settlement convergence that grew in during
review moves to a follow-up PR, together with every blocking finding from the
review rounds.

Pairs with claymorelab/SuperTale#76. **Merge this one first** — EE
depends on the failure-kind contract established here.

## Why split

#311 and its EE counterpart went through five review rounds in about 25 hours
without converging. Each round shipped fixes as new code, which the next round
then reviewed as new code. The two PRs also carry two genuinely different
changes with different risk profiles: the authz retry closure, which is what
the title says, and a settlement convergence design that arrived mid-review and
has its own design doc.

Splitting lets the retry closure land on its own evidence instead of waiting on
the settlement work.

## What is in here

On `staging` the authz round trip has exactly one failure surface. Every way
`admit_model_task` can fail — the admission database being unreachable, a
malformed facts row, any unexpected exception — collapses into
`AuthzError("ORG_AUTHZ_STALE")`, and the consumer's `except Exception` turns all
of it into a `StaleTaskEnvelope`. A transient outage is reported to the user as a
stale envelope, and the task fails permanently with no retry.

This PR splits that surface. `AuthzServiceUnavailable` and `AuthzServiceFault`
are both new here, and they are deliberately siblings under `AuthzError` rather
than one inheriting from the other: `retry_authz_read` retries exactly
`(AuthzServiceUnavailable,)` (`authz_retry.py:62`), so a subclass relationship
would put genuine faults back on the transient retry path. The consumer gets one
arm per kind (`consumer.py:222-224`) plus an explicit `"unknown"` kind for
outcomes matching neither.

This is safe for EE: EE raises these two exceptions and catches neither (four
raise sites in `novelvideo_ee/authz/port.py`), and all three failure kinds still
map onto an existing `_ERROR_BY_REASON` entry:

| failure kind | raised as / code | EE reason | `_ERROR_BY_REASON` |
|---|---|---|---|
| `fault` | `TaskAuthorityFault` / `TASK_AUTHZ_CHECK_FAILED` | `authz_check_failed` | matches |
| `unavailable` | `TaskAuthorityUnavailable` / `TASK_AUTHZ_UNAVAILABLE` | `authz_unavailable_exhausted` | matches |
| `unknown` (new) | as above | `authz_unavailable_exhausted` | matches |

## Verification

Full CE suite, this branch against `origin/staging` (`7369d3e4`):

| | failed | passed | skipped |
|---|---|---|---|
| `origin/staging` | 1 | 3288 | 16 |
| this branch | 1 | 3329 | 16 |

+41 passed, no new failures. The single red,
`test_api_project_summary_paths.py::test_project_summary_counts_from_registered_state_dir`,
is identical on `origin/staging` and is a full-run ordering artifact rather than
a real failure — it passes in isolation on both refs.

## What is deliberately not here

Everything the review rounds flagged as blocking lives in the follow-up PR,
because all of it belongs to the settlement lane: F1 (dropping the
envelope-derived `model_call_credit_policy` turns non-`resolved` outcomes into
per-model-call charging on top of the feature charge) and F3 (the
`ambiguous` / `conflict` branch fails closed with no settlement decision and no
terminalization). The non-blocking findings F2 and F4–F7 travel with them.

None of those findings is reachable from this PR. It does change
`run_core.py` and `ports/usage.py` — `ports/usage.py` gains a
`FeatureCreditSettlementConflict` exception and a
`mark_feature_credit_settlement_for_review` protocol method — but the code the
findings point at is not here: `trusted_billing_metadata`,
`FeatureSettlementResolution` and `FEATURE_SETTLEMENT_RESOLUTION_AMBIGUOUS`
appear in zero source files on this branch, and the five files that read
`model_call_credit_policy` (the `api/routes/*` handlers) are untouched by this
diff.


lywaterman (commented):
## Review — CE #324 (`fix/authz-retry-closure`, head `a8d72456`)

Reviewed against `origin/staging` (merge-base `7369d3e4`), alongside the paired EE PR claymorelab/SuperTale#76 at `d30c553d`.

The split was the right call, and the central change is correct: **the two findings this PR exists to close are genuinely closed.** Two things need resolving before merge, and both are about how this half ships rather than what it computes.

### The split is substantively clean

I verified symbol by symbol that the deferred half is absent:

```
FeatureSettlementResolution         0 occurrences
trusted_billing_metadata            0
_resolve_feature_reservation_id     0
_without_settlement_handles         0
resolve_feature_credit_reservation  0
```

So **F1 and F3 are unreachable from this branch**, as claimed.

But the stated reason is factually wrong, and it matters. The body says:

> None of those findings are reachable from this PR: it does not touch `run_core.py`, `ports/usage.py`, or any settlement path.

All three clauses are false: `run_core.py | 92 +++`, `ports/usage.py | 18 +++`, `ports/local/usage.py | 12 +++`, and the diff adds a **live** settlement call — the `RunningTaskAuthorityIndeterminate` handler in `run_core.py` invokes `get_usage_meter().mark_feature_credit_settlement_for_review(...)`, paired with the equally live `EEUsageMeter.mark_feature_credit_settlement_for_review` added in EE #76.

Right conclusion, wrong reason. A reviewer who takes that sentence at face value skips precisely the one part of this PR that moves money. Suggest rewording to something like: "F1/F3 depend on the trusted resolver, which is not in this PR; this PR does add one live park-in-review settlement path, see §X."

### The central change is correct

`AuthzServiceFault` and `AuthzServiceUnavailable` are now siblings under `AuthzError`, so `authz_retry.py:62`'s default `retryable_error_types=(AuthzServiceUnavailable,)` no longer matches a fault. **That closes N12 from #311.**

`consumer.py:222-226` orders the arms correctly (Fault → Unavailable → AuthzError). Every other CE catch site (`producer.py:131`, `egress_binding.py:79`, `video_generator.py:3378/3454/3491`, `ports/local/tasks.py:161`, the `app.py:274` handler) catches `AuthzError` broadly and is unaffected by the reparenting.

I verified the body's EE claim rather than taking it: EE only ever *raises* these two, at `novelvideo_ee/authz/port.py:172/174/178/188`, and catches neither anywhere in the repo. The `_ERROR_BY_REASON` mapping table checks out.

**The new `"unknown"` kind collapses onto `unavailable`, and that is the right direction.** Previously `except Exception` mapped to `"fault"`, which EE fail-fasts on with zero retries — my #66 finding N5, where a DNS blip killed the user's task on its first attempt. An unknown error now consumes the normal retry budget instead. **That closes N5 too.**

Also done right: the pre-worker refund in `ports/local/tasks.py` that could never fire (the HIGH finding from #311 round 1) is deleted in this branch, not carried forward.

### Blocking

**B1 — the deploy instruction is wrong, and following it loses refunds.**

The body says "Merge this one first", which reads as "CE can ship first". It cannot. This PR removes `billing_metadata` from `RejectedTaskSettlement` and adds `root_task_id`:

```diff
 project_id: str
 requester_user_id: str
+root_task_id: str
 task_type: str
-billing_metadata: Mapping[str, Any]
```

EE `staging`'s `celery_tasks.py:48` reads `settlement.billing_metadata`. On a frozen dataclass without that field this is an `AttributeError`, so **every rejected delivery's refund is skipped for the length of the window.** The reverse is no better: EE #76 against CE `staging` fails at import (`TaskAuthorityFault`, `TaskAuthorityUnavailable`, `root_task_id`) and the worker will not start.

Merge order and deploy order are different things here. Please state explicitly in both PR bodies that this is a lockstep deploy and that neither side may ship alone. (The EE PR has the mirror-image problem; I've filed it there too.)

**B2 — `alembic` now has three heads (EE-side, but it decides whether this pair does anything).**

Filed in detail on EE #76. Summary: `origin/staging` already carries two heads (`0079_org_member_create_roles`, `0082_model_call_payloads`) and EE #76 adds a third (`0078_authz_delivery_terminalizer`), while `scripts/migrate-control-plane.sh` runs `alembic upgrade head` — singular, which errors on multiple heads. If that is the production invocation, the control-plane migration fails, `ee_terminalize_undelivered_task` is never created, and this pair's central deliverable does not work. Needs an answer from whoever owns deploys before either half merges.

### Non-blocking

**F-a. The unsigned-handle money path is fully present, and this PR adds a consumer to it.**

`run_core.py:635`:

```python
run_metadata = {**dict(metadata or {}), **billing_metadata}
feature_reservation_id = _feature_credit_reservation_id(run_metadata)
```

`billing_metadata` comes from `delivery.billing_metadata`, which is not in `_SIGNED_PAYLOAD_FIELDS`. The trusted resolver that fixes this is deferred, so there is no mitigation in this half — that is expected and agreed. What is worth noting is that this PR adds a *new* use of that id (the park-in-review call), so a replayed envelope can now park another tenant's reservation into review. Not a new vector, but one more consumer of an unsigned handle, and the body's "no settlement path" sentence makes it easy to miss.

**F-b. The new park-in-review has no exit, and cannot report its own failure.**

CE writes `source: "task_authz_revalidation_indeterminate"`. EE's `resolve_feature_credit_settlement_reviews` refunds only when `business_outcome ∈ {cancelled, failed, not_delivered}` or `source ∈ _SAFE_INTERRUPTION_SOURCES`; this source is in neither and carries no `business_outcome`, so the hourly pass defers it indefinitely. That much matches the documented design in `frontend-spec.md` §12.2 ("running 只进 review；不自动退款、不确认"), resolvable through `resolve_awaiting_credit_settlement`.

The bug is on the EE side and I've filed it there: `ports_bootstrap.py:376` returns a hard-coded success dict even when the guarded `UPDATE` matched zero rows or settlement is disabled — so CE's `except Exception: logger.error(...)` here can never fire, and a park that silently no-ops is indistinguishable from one that worked.

**F-c. Minor, carried over.** `SettlementIntentResult.retryable` is still CE-dead — both call sites discard the return value (EE-only). The new `error_id` is a fresh uuid that is never persisted, returned, or attached to the settlement record, so it correlates with nothing, and `reservation_id` is still absent from the failure `extra` (my #311 N9).

### Deferred by design — not re-raising these here

For the record, so they don't get re-litigated on this PR: F1 (dropping `model_call_credit_policy` turns non-`resolved` outcomes into per-call charging), F3 (`ambiguous`/`conflict` fails closed with no settlement decision and no terminalization), F2, F4–F7, and the whole model review/awaiting quarantine. All confirmed absent from this branch. I'll review them on the follow-up PR, together with a check that the S1–S5 plan's items arrived as designed.

Handanhhhy:
Addressed the review shipping blockers in the paired Lane A update. The body now distinguishes merge order from lockstep deployment and accurately calls out the live park-in-review settlement path. EE #76 head 7a4e9cfb now has bounded, independently counted marker-only dispatch retries and a single-head migration graph. Please re-review the pair together.

lywaterman (commented):
## Follow-up — body corrections confirmed; blocked on EE #76

The body edit addresses what I raised. It now separates merge order from deployment ("Deployment is lockstep: neither ref may ship or run against the old counterpart"), states plainly that post-accept authorization uncertainty parks the feature reservation in review and that **this is a live settlement path**, and replaces the earlier "does not touch `run_core.py`, `ports/usage.py`, or any settlement path" with an accurate description of what is and is not present. The F1/F3 deferral rationale is now stated in terms of `FeatureSettlementResolution` / `trusted_billing_metadata` / the trusted resolver being absent, which is what I verified symbol by symbol. Nothing further from me on the body.

Lockstep deployment is settled operationally, so I am not raising deploy ordering again.

The code on this branch is unchanged this round (`a8d72456`), so my previous review stands as-is: the split is substantively clean, N12 and the #66 N5 fail-fast problem are both closed, and the pre-worker pseudo-refund removal is correct.

**This half is still gated on EE #76.** I filed a follow-up there: B1 (three alembic heads) is fixed, but B2 is not, and the current EE head regresses a path that previously worked. Summary, since it decides whether this pair functions at all:

EE's `self.retry(kwargs={"envelope": ...})` is called without `args`, while the producer dispatches positionally (`task_backend.py:633`, `args=[envelope]`). Celery's `signature_from_request` keeps `args = request.args`, `apply_async`'s typing check then raises `TypeError: got multiple values for argument 'envelope'`, and `Task.retry` converts that to `Reject(exc, requeue=False)` — which, under `task_acks_late=True`, drops the message. Every retry path in `run_project_task` is now one-shot: a single authz 503 that used to retry is rejected and dropped, with no retry, no terminalization dispatch and no refund. The fix is small (pass `args` explicitly), but it needs to land before this pair is deployable.

One item from my previous review that is worth taking on the EE side while that fix is made, because it pairs directly with the body change here: `ports_bootstrap.py:376`'s `mark_feature_credit_settlement_for_review` returns a hard-coded success dict even when the guarded `UPDATE` matched zero rows or settlement is disabled. This branch's `RunningTaskAuthorityIndeterminate` handler wraps that call in `except Exception: logger.error(...)`, so with the EE side always reporting success, a park that silently no-ops is indistinguishable from one that worked — on the path the body now correctly labels as live settlement.

lywaterman (commented):
## Scope for closing this pair

Confirming where this half stands, so nothing is left ambiguous.

**CE #324 has no blocking findings.** The split is clean, `AuthzServiceFault` reparenting closes #311's N12, the `"unknown"` failure kind closes #66's N5, the unreachable pre-worker pseudo-refund is correctly removed, and the body now describes the contents accurately. From my side this half is ready to merge.

**It is not ready to deploy, because the pair is gated on EE claymorelab/SuperTale#76 B2.** EE's `self.retry(kwargs=...)` is called without `args` while the producer dispatches positionally, so every retry in `run_project_task` is rejected and dropped rather than retried. Filed there with the fix; it is a small change.

**Out of scope for this pair by agreement**, recorded so it is not re-litigated: EE's N8 and N9, plus everything in the settlement lane — F-a (the reservation id still derives from unsigned `delivery.billing_metadata`, since the trusted resolver is deferred), park-in-review having no automatic exit, and F1, F3, F2, F4–F7 and the model review quarantine. I'll review those on the follow-up PR, together with a check that the S1–S5 plan's items arrived as designed.

**One item I'd still like taken with the EE B2 fix**, because it pairs with the body change here rather than with the settlement lane: `ports_bootstrap.py:376`'s `mark_feature_credit_settlement_for_review` returns a hard-coded success dict even when the guarded `UPDATE` matched zero rows or settlement is disabled. This branch wraps that call in `except Exception: logger.error(...)` on the `RunningTaskAuthorityIndeterminate` path — the path the body now correctly labels a live settlement path — so while EE always reports success, that error handler is unreachable by construction and a silent no-op park looks identical to a successful one.

Nothing else from me on this PR until EE B2 lands.

PR #326, opened 2026-08-20
fix: converge paid execution settlement

Handanhhhy:
## Summary

- 从可信 feature settlement snapshot 重建 billing metadata 与 model-call credit policy
- 对非 resolved 结果 fail closed，避免静默回退到 per-model-call charging 造成重复计费
- 为 ambiguous/conflict 产生可签名、可重试的 settlement-only terminalization marker
- marker dispatch retry 不重新执行授权检查、业务任务或 provider 请求

## Verification

- Lane S 定向回归：62 passed
- Black、Ruff、git diff --check 通过
- 已核对当前分支包含最新 origin/staging

## Release contract

本 PR 目标为 staging，对应 EE PR：claymorelab/SuperTale#78。CE/EE 必须 lockstep merge/deploy，不得单独上线。


lywaterman (commented):
## Review — CE #326 (`fix/paid-execution-settlement`, head `8fdc9062`)

Reviewed against `origin/staging` (merge-base `2294dbb8`), alongside EE claymorelab/SuperTale#78 at `f934c023`.

**F3 is closed. F1 is not — it survives on the `not_applicable` branch, and it is a regression against `staging`.**

A note on method: I could not get tests run for this pair, so the verification below is from reading code and tracing the runtime path across both repos. I am not confirming or disputing the "Lane S 定向回归: 62 passed" figure.

### F3 — closed

`FeatureSettlementResolutionRejected` now carries `code = f"FEATURE_SETTLEMENT_RESOLUTION_{outcome.upper()}"` (`ports/usage.py:83`) and the handler returns `{"failed": True, "error_code": exc.code}` (`run_core.py:728`). On the EE side `_SETTLEMENT_RESOLUTION_REASON_BY_ERROR_CODE` (`celery_tasks.py:32-34`) now maps all three codes — `..._AMBIGUOUS`, `..._CONFLICT`, `..._FAILED` — to terminalization reasons, and `:294-300` dispatches the signed marker on any of them. So `ambiguous` and `conflict` now reach a settlement decision instead of stranding the reservation. That was the substance of F3.

### F1 — still open on `not_applicable`

The PR body says "对非 resolved 结果 fail closed". In the code only `ambiguous` and `conflict` fail closed; `not_applicable` returns normally:

```python
# run_core.py:251-257
if resolution.outcome == "resolved":
    resolution.trusted_billing_metadata()
    return resolution
if resolution.outcome == "not_applicable":
    return resolution
raise FeatureSettlementResolutionRejected(resolution.outcome)
```

From there the original F1 chain is intact:

1. `ports/usage.py:42-46` — `__post_init__` forbids `reservation_id` / `feature_key` / `model_call_credit_policy` on any non-`resolved` outcome, so `trusted_billing_metadata()` is necessarily `{}` for `not_applicable`.
2. `run_core.py:769-771` — `if trusted_billing_metadata:` is false, so `envelope["billing_metadata"]` is never set; `:826` passes `{}` into `_set_project_task_metrics_context`.
3. EE `metrics_emit.py:559` reads `model_call_credit_policy` only from `get_billing_metadata_context()`, and the short-circuit at `:591` fires only on `== MODEL_CALL_CREDIT_POLICY_FEATURE_INCLUDED`. An empty value falls through to `reserve_model_call_credits` — per-model-call charging.

**This is a regression against `staging`, not a pre-existing gap.** EE `task_backend.py:497-511` sets the policy *before* the reservation is attempted, gated only on `feature_spec is not None`:

```python
if feature_spec is not None:
    ...
    billing_metadata["model_call_credit_policy"] = feature_spec.model_call_credit_policy
    billing_metadata["feature_key"] = feature_spec.feature_key
    try:
        reservation = await reserve_feature_start_credits(...)
```

So on `staging` the policy reached the context even when no reservation was created, and the short-circuit suppressed per-call charging. This PR stops trusting the envelope — correctly — but replaces it with a snapshot that is empty in exactly the "no resolvable reservation" case.

**Reachability.** `feature_billing.py:150` carries an `allow_free=True` spec; a zero-cost feature creates no reservation and therefore no `awaiting` outbox row, so the resolver returns `not_applicable` and the feature's model calls start consuming per-call credits. Redelivery after the row has been settled or reaped lands in the same branch, as does any drift in the resolver's ledger join predicates — `not_applicable` doubles as "a reservation exists but a predicate did not match", which was already noted on the EE side.

One path I checked and can exclude: a failed reservation raises at `task_backend.py:613-614` and the task is never enqueued, so that does not reach `not_applicable`.

**Suggested fix.** The cleanest shape keeps the policy out of the `resolved` gate entirely, since it is a compile-time property of the feature spec rather than a fact about a specific reservation: derive it from the resolved `feature_key` when the outcome is `resolved`, and fall back to the signed `identity.task_type` otherwise, failing closed only when neither yields a spec. `feature_billing_spec()` already accepts either key (`feature_billing.py:425-427`), so both lookups work. Deriving it *only* from `task_type` would not be equivalent — `_payload_feature_billing_spec` resolves `explicit_feature_key or task_type` (`task_backend.py:156-158`) and several routes pass an explicit `billing.feature_key` that differs from the task type, so the two can select different specs.

### F7 — partially closed

`__post_init__` requires only `reservation_id` when `outcome == "resolved"` (`ports/usage.py:40-41`); `feature_key` and `model_call_credit_policy` are still optional. In practice EE always populates both, because `ports_bootstrap.py:205-207` raises when the spec lookup fails — so a partial `resolved` snapshot should not occur today. But the invariant itself does not enforce it, which was the point of the finding.

### N9 — still open

The settlement failure handlers (`run_core.py:362`, `:395`, `:434`) still emit `safe_error_type` plus a freshly minted `error_id` that is never persisted, returned, or attached to the settlement record, with no `reservation_id` in `extra` and no `exc_info`. A failed confirm or refund still cannot be traced to the affected credit row. Adding `reservation_id` restores correlation without reintroducing adapter text.

### Cross-repo

EE #78 closes its side cleanly — the model-review evidence is now keyed on the model reservation id against a column that `0085` actually adds and that `metrics_emit.py:643` actually writes, both confirm and refund branches are reachable, the defer path escalates to an operator alert, and the quarantine scoping that keeps the resolver away from CE's legitimate post-acceptance refunds is still in place. Details on that PR.

Handanhhhy:
@lywaterman Lane S review fixes are now at `8a032405`.

This update closes the remaining CE-side settlement gaps:
- reservation-free/zero-cost features now retain trusted `feature_key` and `model_call_credit_policy`, so `not_applicable` cannot silently fall back to per-model-call charging;
- resolved snapshots are complete at construction time; EE durable lookup uses an internal intermediate type before policy hydration;
- ambiguous/conflict paths emit failed-task metrics;
- provider exceptions carry only normalized HTTP status evidence for conservative model settlement decisions.

The previously missing DCO trailers were repaired with a tree-identical history rewrite; the PR head now has matching sign-offs throughout.

Verification: CE targeted settlement/authz/timeout/instrumentation suite: `91 passed`. Please re-review when convenient.

lywaterman (commented):
## Follow-up review — `8a032405`

**F1 and F7 are closed, and F1 is closed the right way.** One new blocking issue, which lives on the EE side; N9 is still open.

Method note: delegated review agents failed repeatedly on this pair, so everything below is from reading code across both repos. I could not run the suite, so I am not confirming or disputing the "91 passed" figure.

### F1 — closed

Two changes together do it. `ports/usage.py`'s `__post_init__` now lets a `not_applicable` resolution carry `feature_key` and `model_call_credit_policy` (both or neither), and `trusted_billing_metadata()` emits for both `resolved` and `not_applicable`, filtering empty values — so `not_applicable` yields the policy and feature key without a reservation id. On the EE side the resolver rehydrates from the spec registry, keyed on `resolution.feature_key` when resolved and on the signed `identity.task_type` otherwise.

That is the shape I suggested, and it matters that they split it by outcome rather than keying everything off `task_type`: the resolved path still uses the value the reservation was actually made under, so a payload that selected a different spec via an explicit `billing.feature_key` cannot drift from what was reserved.

Net effect: `model_call_credit_policy` reaches `_BILLING_METADATA_CTX` for reservation-free features, `metrics_emit.py:591`'s `feature_included` short-circuit fires, and the per-model-call fallback no longer triggers. The `allow_free=True` spec I flagged as the concrete reachable case is `freezone_text_translate`, which is itself a registry key, so the `task_type` lookup finds it.

### F7 — closed

`__post_init__` now requires a complete snapshot on `resolved`:

```python
if self.outcome == "resolved":
    if not clean_reservation_id:
        raise ValueError("resolved settlement requires reservation_id")
    if not clean_feature_key or not clean_model_call_credit_policy:
        raise ValueError("resolved settlement requires a complete billing snapshot")
```

A partial `resolved` snapshot can no longer fail open toward charging.

### N10 — closed

`:732-738` emits `_emit_project_task_metrics(..., outcome="failed")` before returning on the `FeatureSettlementResolutionRejected` path, so `ambiguous` and `conflict` are no longer invisible in task-outcome metrics.

### Blocking — image-edit tasks hard-fail on `not_applicable` (fix belongs in EE #78)

`feature_billing.py:434` defines a third category of task type that is neither registered nor exempt:

```python
# Shared image-edit runners choose one of the registered image feature keys
# from the request payload, so the task name itself intentionally has no spec.
EXPLICIT_FEATURE_BILLING_TASK_TYPES = frozenset({"freezone_edit", "freezone_mask_edit"})
```

The set is referenced only in `feature_billing.py` and a registry test. The new resolver's `not_applicable` branch keys on `identity.task_type` alone, finds no spec, finds the task is not exempt, and raises `RuntimeError("feature settlement has no billing spec")` (`ports_bootstrap.py:216`).

That lands in this PR's generic `except Exception` at `run_core.py:729`, becomes `FEATURE_SETTLEMENT_RESOLUTION_FAILED`, and the task hard-fails where it previously ran. Reachable whenever an image-edit task hits `not_applicable` — redelivery after the outbox row was settled or reaped, ledger join-predicate drift, or a zero-cost edit. Redelivery is routine; it is why `acks_late` and the idempotent terminalizer exist.

Worth noting for the fix: `_SIGNED_PAYLOAD_FIELDS` is `{"episode", "beat_num", "scope", "queue_kind", "payload"}`, so `payload` is HMAC-covered and `payload["billing"]["feature_key"]` is already trusted data. The resolver just does not look at it. Mirroring the enqueue-side `explicit_feature_key or task_type` rule for the `not_applicable` branch closes this without widening the trust boundary. Filed in detail on EE #78, since the raise and the missing category check are both there.

### N9 — still open

The settlement failure handlers (`run_core.py:363`, `:396`, `:438`) still emit `safe_error_type` plus a freshly minted `error_id` that is never persisted, returned, or attached to the settlement record, with no `reservation_id` in `extra` and no `exc_info`. A failed confirm or refund still cannot be traced to the affected credit row. Adding `reservation_id` restores correlation without reintroducing adapter text — it is an internal ledger id, not sensitive.

### Cross-repo

EE #78 closes all six of its deferred findings, and I verified each mechanism rather than the claim: the model-review evidence is now keyed on the model reservation id against a column `0085` actually adds and `metrics_emit.py:643` actually writes; both confirm and refund branches are reachable; the defer path escalates to an operator alert; the quarantine scoping that keeps the resolver away from CE's legitimate post-acceptance refunds is intact; and `0085` leaves Lane A's `0078` untouched at a single head.

PR #323, opened 2026-08-20
feat(chat): per-turn gateway credentials and the BrainClaw evidence plane

lywaterman:
Brings the per-turn credential and the BrainClaw evidence plane onto
`freezone-canvas`. Deployed and collecting in the test environment; the full
chain canary runs 123 PASS / 0 FAIL.

## What this changes for everyone, not only BrainClaw users

Three of these are not opt-in features, and that is the part worth reviewing
closely:

- every Hermes worker now carries the per-turn credential latch;
- a latched worker may reach only its configured gateway;
- the workspace no longer holds a gateway key on disk.

A deployment pointing Hermes at its own OpenAI-compatible endpoint never asked
for any of that, so `tests/test_brainclaw_non_regression.py` states what such a
deployment is entitled to: an ordinary turn still authenticates with the
configured key, no capability is required to take a turn, and an unconfigured
issuer is not counted as a failure. The confinement rule is *only the
configured gateway*, not *only BrainClaw's*, compared by origin — a BYO
endpoint is its own gateway.

The honest consequence is an install requirement rather than a setting:
**this DramaClaw runs only with the paired Hermes fork.** A stock Hermes drops
the `_meta` extension the credential travels in, so every turn would fail closed
and surface as a connection error. `require_hermes_fork` refuses before a worker
is spawned, probing behaviour rather than version — the fork keeps upstream's
version string, so `hermes --version` cannot tell them apart.

There is no legacy mode. One existed briefly and was worse than none: it put the
real key back on disk while the worker kept the placeholder and the latch, so a
"legacy" deployment carried the old design's exposure and the new design's
behaviour at once.

## The evidence plane

DramaClaw signs a per-turn capability; the gateway verifies it, mints a
request-bound Control Context, and assigns a durable checkpoint ordinal;
BrainClaw verifies that and records it. Three keys, three roles, no side holding
one it must not have — the grouping key never leaves DramaClaw, because it
derives the opaque identifiers and a copy elsewhere turns a verifier into a
minter.

Reason counters on all three sides, so a chain that quietly stops producing
evidence is distinguishable from one with no traffic.

## Defects this found and fixed

- **`_meta` never reached the handler.** The ACP router splats it into keyword
  arguments, so `kwargs.get("_meta")` was always `None` and the credential was
  never bound. Twenty-eight tests passed throughout, because every one bound the
  ContextVar directly and none crossed that boundary.
- **The turn identity could outlive its turn.** Capability and credential were
[...]

PR #328, opened 2026-08-20
fix(deploy): install the Hermes fork in local setup too

lywaterman:
Follow-up to #323. That PR moved the *image* to installing Hermes from this
project's fork; `scripts/setup-hermes.sh` was left installing
`hermes-agent[acp]==<version>` from PyPI, so a developer following the
documented setup ends up with a different Hermes than the image runs.

That difference is not cosmetic. A PyPI release keeps the same version string
as the fork and then drops the `_meta` extension the per-turn credential
travels in — so every turn fails closed, the OpenAI SDK wraps the refusal as a
connection error and retries three times, and the symptom arrives fourteen
seconds later pointing at the network while the gateway has seen no request at
all. It took a long time to find the first time.

## What changes

**The script installs the fork**, by cloning and installing editable — mirroring
what the Dockerfile does. `uv tool install` cannot be used here at all:
upstream refuses to build a wheel or sdist on purpose ("distributed via the
shell installer, Docker image, or Nix"), which is also why the original script
could only reach PyPI.

**It verifies behaviour instead of a version.** Comparing `hermes --version`
against a pinned file cannot tell the fork from stock, because the fork keeps
upstream's version string. The check now asks the question the deployment
actually depends on — does `_meta` survive the ACP router — and its `--check`
mode distinguishes "nothing installed" from "something installed that is not
the fork", since those need different fixes.

**`.hermes-version` is deleted** along with its last reader. The runtime gate
that consumed it went in #323; which upstream release the fork carries is a
property of the branch now, not a number to keep in step. Two compose comments
and one docstring referenced it and are updated.

Overridable with `HERMES_REPO`, `HERMES_REF` and `HERMES_SOURCE_DIR`, matching
the Dockerfile's build args.

## Checks

- `--check` correctly reports a non-fork Hermes in an environment that has one
- the image still builds, and `verify_hermes_fork.py` passes inside it
- `Dockerfile` no longer COPYs the deleted file, which would have failed the
  build


vkiki (commented):
感谢 PR #328 的 Hermes fork 本地安装修复方向。我们复核后认为方向正确，但合并前建议补充以下修正：

1. setup-hermes.sh 当前用系统 python3 检查 fork 模块，但实际运行的 Hermes 由 PATH/HERMES_CLI_PATH 决定，可能出现“检查的是 fork、实际启动的是旧 PyPI CLI”的误判。建议检查实际 hermes 可执行文件对应的解释器和模块。
2. fork_is_installed() 没有先确认 hermes CLI 存在；仅模块导入成功时可能误报安装成功。建议同时校验 command -v hermes 和 hermes --version。
3. 本地使用 uv pip install --system -e，可能写入系统 Python，普通开发环境会遇到权限或 Python 环境不一致问题。建议使用项目或用户隔离环境，并明确 HERMES_CLI_PATH。
4. Dockerfile 已改为 HERMES_REPO/HERMES_REF，但 Compose 仍传递旧的 HERMES_INSTALL_SPEC，导致无法按文档通过 Compose 固定 commit，默认分支也不可复现。建议同步 Compose build args，并优先使用固定 SHA。
5. 建议增加安装脚本测试：PyPI Hermes 与 fork 混装、CLI 路径不一致、--check、固定 SHA，以及实际验证 _meta 通过 ACP 路由。

在这些问题处理前，我们建议暂不合并；处理后可以再次复审。

PR #330, opened 2026-08-20
fix: close scene pano model injection

Handanhhhy:
## Summary

- restrict scene panorama API inputs to server-controlled model settings
- reject untrusted raw model passthrough in the worker parser
- preserve server-authorized dynamic catalog models through a signed task authority envelope
- remove the unused frontend free-form scene panorama submission helper

## Verification

- `uv run pytest -q -p no:cacheprovider tests/test_freezone_image_backend.py tests/test_p0g4l_scene_360_subprocess_egress.py tests/test_stage_asset_scene_360_credit.py tests/test_stage_asset_scene_360_provider.py` (228 passed)
- focused related suite (355 passed)
- Ruff checks passed
- `git diff --check` passed

## Scope

OI-51 P0 partial: input convergence and parser rejection of raw passthrough.

### DIFF
--- a/frontend/src/api/ops.ts
+++ b/frontend/src/api/ops.ts
@@ -1772,8 +1772,7 @@ export interface FreezoneScene360Payload {
  * **单图 360 (simple pipeline)** — 老路径,只把一张图当 reference 走通用
  * 图编辑生成 panorama。不做 master/reverse overlap 分析、空间合约、缝合,
  * 适合 freezone 自由画布上 \"拿任一张图试试 360\" 的快速生成。
- * Asset-scoped scene 360 generation uses
- * {@link submitFreezoneScene360FromMaster} (复杂工作流),不是这条。
+ * Asset-scoped scene 360 generation uses the scenes asset API (复杂工作流),不是这条。
  */
 export async function submitFreezoneScene360(
   project: string,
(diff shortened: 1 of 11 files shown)


lywaterman (commented):
## Review — #330 (`fix/oi51-scene-pano-input`, head `09867285`)

Reviewed against `origin/staging` (merge-base `1ccb573e`). The security design holds up — I traced the trust boundary rather than taking the summary's word for it. One functional point worth addressing, filed as a suggestion rather than a blocker since it does not weaken the fix.

### The trust boundary is sound

Four layers, each of which I checked:

**1. The untrusted route no longer accepts model input.** `ScenePanoGenerateRequest` gains `extra="forbid"` and drops `style`, `provider`, `model`, `image_size`, `quality` and `timeout_seconds` entirely, so `generate_scene_pano` has nothing to forward. `scenes.py` now uses the server-side `_project_style(username, project_name)` and a literal `timeout_seconds=1800`, which matches the old schema default — no behaviour change there.

**2. Billing identity and model authority are properly separated.** `catalog_id` may still fall back to `body.catalog_id`, but `execution_catalog_id` comes only from `_catalog_entry_id(catalog_entry)` — the server-resolved entry — and never from the body. The comment at `freezone.py:4705-4707` states the distinction, and the code matches it.

**3. The authority envelope rides inside the signed payload.** `scene_360_model_authority` is added to the `payload=` kwarg, and `payload` is in `_SIGNED_PAYLOAD_FIELDS`, so it is HMAC-covered rather than travelling as unsigned metadata.

**4. The runner's capability marker is real.** The guard is `type(envelope) is not TrustedRunnerEnvelope`. I checked the type: `egress_context.py:14` defines it as a `dict` subclass, and the only construction site in the repo is `run_core.py:695`, inside the verified task core after `consume()` has run. Using `type(...) is` rather than `isinstance` also means a subclass cannot spoof it. So a plain dict from any other caller fails closed with `InvalidTaskEnvelope`.

The parsing guards are appropriately strict: exact key set `{"kind", "catalog_id", "provider", "model"}`, `kind == "catalog"`, and `type(value) is str and value.strip()` on each field, with any mismatch raising `InvalidTaskEnvelope`.

One note on reasoning rather than code: the runner cross-checks `authority_provider`/`authority_model` against `params["provider"]`/`params["model"]`, but both sides live in the same signed payload, so that comparison guards against internal inconsistency, not forgery — anyone able to forge one could forge both. The actual boundary is the signature plus `TrustedRunnerEnvelope`. Worth keeping straight so the check is not later mistaken for the security property and relaxed.

### Suggestion — the rejection lands in the worker rather than at the API

`resolve_scene_360_image_model` now raises `Scene360ImageModelSelectionError` where it previously fell through to `return resolved_model`. That is the right hardening.

But the legacy route still builds the model as:

```python
# freezone.py:4697
"model": catalog_model or body.model or FREEZONE_DEFAULT_IMAGE_MODEL,
```

When `_resolve_catalog_request` resolves no entry, `catalog_entry` is `None`, so `execution_catalog_id` is `None`, no `scene_360_model_authority` is attached, and the runner calls `run_scene_360` with `model_authority=None`. `resolve_scene_360_image_model` then raises for any `body.model` that is not in `IMAGE_GENERATION_SELECTIONS`.

The injection is correctly blocked — that is the point of the change. The issue is where it fails: after the request has been accepted, the task queued, and the feature credit reserved. The user sees a failed task and the system spends a settlement round-trip on it, instead of getting a 4xx at submit time.

Applying the same check at the route — reject when the catalog resolves no entry and the requested model is not registered — would turn this into a clean rejection and keep unauthorized models out of the queue entirely.

### Smaller notes

- The `selection is not None` branch now raises when the registered selection's provider does not match the resolved provider, instead of falling through to a raw passthrough. Correct direction, and it closes the same hole as the unknown-model case.
- Removing the free-form submission helper from `ops.ts` (-71) matches the backend tightening, so no client is left calling a route that would now reject it.
- Test coverage looks proportionate: contract, backend, egress, provider and credit paths all get cases. I did not run them, so I am not confirming the 228 / 355 figures.

PR #335, opened 2026-08-20
feat: default UI/Hermes chat language to English

patilswapnilv:
## Summary
- Default new-user UI locale to English (`i18n` / app-store / release notifications); persisted `zh` users keep Chinese
- Pass UI `language` on chat WebSocket turns and inject locale-aware Hermes reply contracts (SOUL/MEMORY, plugin instructions, stop/error strings)
- Add EN/ZH local-LLM (Ollama / hybrid) guide and cross-links from docs/README

## Test plan
- [ ] Clear site data → UI opens in English; switch account menu language → UI follows
- [ ] Hermes chat with `language=en` replies in English; with `language=zh` replies in Chinese
- [ ] `uv run pytest tests/test_chat_reply_language.py tests/test_hermes_workspace.py`
- [ ] Frontend: `vitest` for release-notifications + app-store language defaults

### DIFF
--- a/.hermes/plugins/dramaclaw/__init__.py
+++ b/.hermes/plugins/dramaclaw/__init__.py
@@ -125,8 +125,9 @@ def _with_chat_error_hints(value: Any) -> Any:
         result.setdefault(
             "agent_instruction",
             (
-                "Reply to the user with chat_error in natural Chinese. Make clear the audio task "
-                "was not started. Tell the user they can go to 虾塘 to upload or record the missing "
+                "Reply to the user with chat_error in the user's UI language "
+                "(default English). Make clear the audio task "
+                "was not started. Tell the user they can go to Freezone (虾塘) to upload or record the missing "
                 "voice lines, then continue. Do not start another tool in this turn."
             ),
         )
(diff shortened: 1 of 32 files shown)


NewYuee:
Thanks for working on the English-default experience. I reviewed the current commit and found three issues that should be addressed before merge:

1. **English re-ingest confirmation cannot complete.** The backend asks English users to reply `overwrite`, then `confirm` or `continue`, but the frontend confirmation helpers only accept `覆盖`, `确定`, and `继续`. An English reply is therefore treated as cancellation. Please make the confirmation parsing locale-aware or safely accept both English and Chinese tokens, and align the hard-coded confirmation context with the same rules.

2. **English script-creation requests bypass the upload-guidance guard.** The new English response template is selected correctly, but `_SCRIPT_CREATION_REQUEST_RE`, `_STYLE_SHORT_DRAMA_REQUEST_RE`, and `_CONTINUE_PIPELINE_RE` still only recognize Chinese phrases. For example, `Write me a short drama script` currently returns no guidance and falls through to the agent. The new test uses a Chinese request with an English response language, so it does not cover this case. Please add English request detection and regression tests using actual English input.

3. **Existing backend tests regress.** Running the focused suite produced `2 failed, 99 passed`. The two re-ingest tests call `stream_assistant_reply()` without a language and still assert Chinese output, while the new default is English. Please either pass `language="zh"` where the test intends to cover Chinese behavior or update the expected default output.

Command used:

```bash
PYTHONPATH=src python -m pytest \
  tests/test_chat_reply_language.py \
  tests/test_hermes_workspace.py \
  tests/test_chat_service_user_agent_scope.py -q
```

`git diff --check` passes.

PR #367, opened 2026-08-24
feat: give structured characters the face, role and build they never had

lywaterman:
feat: give structured characters the face, role and build they never had

structured_v1 characters land with an empty 面部提示词, 角色定位 and 身形.
This is not a regression: `git log -S face_prompt` over the structured
extraction path returns nothing since 3c6eb619, its first commit. The stage
that writes those fields was never carried over.

Legacy does it in one call. `CharacterEnrichment` asks a model for nine
fields at once — who the character is *and* what they look like — off the
graph context plus the script's character bible. Structured extraction only
carried over the first half. Its contract is evidence-bound: every field a
candidate reports has to appear verbatim in the span it came from, which is
what stops an invented character from reaching the table. A face prompt, a
role and a build are none of them quotable, so they could not ride along in
CharacterCandidate without relaxing the guard that makes extraction worth
trusting.

Scenes never had this problem. `enrich_scene_environments_batched` has been
there from the beginning, which is exactly why scene builds produce usable
prompts and character builds did not.

The consequence is not cosmetic. character_image.py refuses outright:

    if not face_prompt:
        raise RuntimeError("请先设置面部特征 (face_prompt)")

So 生成新图 on any structured project's character card fails.

## The stage

Runs after adjudication has settled the cast, takes only names that already
survived verification, and is allowed to write what the source implies
rather than what it states. Two contracts, kept apart.

- role, is_main, age_group, body_type, face_prompt; fish_voice_id derived
- cached in story_analysis_item_cache under `character_appearance`, keyed on
  name, aliases, gender, description, the quotes the prompt actually sends,
  the synopsis, and a contract version
- an answer with no face prompt is neither stored nor published. Caching one
  would make the omission permanent — the trap is_cacheable_scene_prompt
  already guards against on the scene side
- age bands are coerced to the four the column accepts, because
  get_fish_voice_id and identity planning both derive from it and an unknown
  value would silently disable them rather than fail
- at most one narrator, resolved in the caller's order: batches run
  concurrently, so "whoever answered first" is not a stable answer
- a failed batch degrades. Losing a face must not lose the character

The screenplay's pre-scene block goes in whole. It names hair, build and
bearing outright, it is what legacy fed the same question, and scene
enrichment has taken it as `synopsis` all along.

## Existing characters

Published through `_publish_characters`, so a run replayed from its stored
cast is enriched too. Without that, a project built before this stage
existed could never acquire a face without deleting its characters first.

An existing character is an asset fact and a rebuild leaves it alone —
except for fields it never had. Repair is field by field: a role written by
hand survives supplying the face beside it. The age band has no empty state,
so an unbound voice is what distinguishes a band nobody chose from one
somebody did; where nothing is bound, the band and the voice it selects are
written together and cannot end up disagreeing.

## A voice bug on the way past

    gender_key = "female" if "女" in gender else "male"

legacy's prompt asks for 男/女. The structured extractor's asks for
male/female. So every woman in every structured project has been getting a
male voice — and not only here: identity_planner.py:402 calls the same
function with the same value. Both spellings now resolve. Legacy's inputs go
down the same branch they always did.

## Tests

16 new. Six were checked against the unfixed tree and fail there: a new
character publishes with its face and voice, a rebuild fills a face an
existing character never had, a bound voice keeps its age band, the voice
lookup reads both genders, and the character bible reaches the prompt.

The rest cover the guards: a faceless answer is dropped rather than cached,
an unusable age band falls back, a cached appearance is never re-sent, an
edited description / quote / bible retires the stored answer, only one
narrator survives a batch split, a failed call still publishes the cast, and
a face the user wrote is never overwritten.

ruff clean. 3838 passed, 16 skipped.

## Not in this change

Legacy's prompt also excludes animals, monsters and robots; structured
extraction's does not. That belongs to the extraction prompt, not to this
stage, and it is a fail-closed guard — it deletes real characters when it is
wrong — so it wants its own evaluation.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/src/novelvideo/structured_builders.py
+++ b/src/novelvideo/structured_builders.py
@@ -198,6 +198,72 @@ async def build_characters_structured(
     )
 
 
+
+def _source_synopsis(store: Any) -> str:
+    """The screenplay block before the first scene, or "" for prose.
+
+    Read here rather than threaded in from the build, because a cast replayed
+    from a stored run reaches publication without the source text in hand and
+    must still be described from the same input a fresh build used.
+    """
+    from novelvideo.cognee.script_parser import extract_synopsis
+
+    try:
+        return extract_synopsis(require_imported_novel(store.project_dir))
+    except Exception:  # noqa: BLE001 - a missing source only costs context
+        return ""
+
+
+def _apply_appearance(character: Any, appearance: Any, voice_for: Callable) -> None:
+    """Write an appearance answer onto a character about to be created."""
+    if appearance is None:
+        return
+    character.role = appearance.role
+    character.is_main = appearance.is_main
+    character.age_group = appearance.age_group
+    character.body_type = appearance.body_type
+    character.face_prompt = appearance.face_prompt
+    character.fish_voice_id = voice_for(appearance.age_group, character.gender)
+
+
+async def _repair_missing_appearances(
+    store: Any,
+    added: set,
+    appearances: dict,
+    voice_for: Callable,
+) -> int:
+    """Fill only the appearance fields an existing character still lacks.
+
+    Field by field rather than wholesale: a character may have been given a
+    role by hand and left without a face, and overwriting the first to supply
+    the second would lose an edit the user made deliberately.
+    """
+    repaired = 0
+    for name, appearance in appearances.items():
+        if name in added:
+            continue
+        existing = store.get_character(name)
+        if existing is None or str(existing.face_prompt or "").strip():
+            continue
+        updates: dict[str, Any] = {"face_prompt": appearance.face_prompt}
+        if not str(existing.role or "").strip():
+            updates["role"] = appearance.role
+        if not str(existing.body_type or "").strip():
+            updates["body_type"] = appearance.body_type
+        if not str(existing.fish_voice_id or "").strip():
+            # The age band has no empty state — it defaults to "youth" — so an
+            # unbound voice is what distinguishes a band nobody chose from one
+            # somebody did. Where nothing is bound, the band and the voice it
+            # selects are written together, so they cannot end up disagreeing.
+            updates["age_group"] = appearance.age_group
+            updates["fish_voice_id"] = voice_for(
+                appearance.age_group, existing.gender
+            )
+        await store.update_character(name, **updates)
+        repaired += 1
+    return repaired
+
+
 async def _publish_characters(
     store: Any,
     merged: list,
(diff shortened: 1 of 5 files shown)


src/novelvideo/structured_builders.py
```diff
@@ -198,6 +198,72 @@ async def finish(status_error: tuple[str, str]) -> None:
...
+    the second would lose an edit the user made deliberately.
+    """
+    repaired = 0
+    for name, appearance in appearances.items():
+        if name in added:
+            continue
+        existing = store.get_character(name)
+        if existing is None or str(existing.face_prompt or "").strip():
+            continue
+        updates: dict[str, Any] = {"face_prompt": appearance.face_prompt}
```
Handanhhhy: [P1] Please publish `is_main` when repairing pre-stage characters. `_apply_appearance` sets it for newly inserted rows, but this repair path never does, so an existing structured project remains with every character at `is_main=False` even when the appearance result identifies its narrator. That later makes first-person narrator resolution fail with `未找到解说主角（is_main=True）`. Minimal red test: create an existing faceless character, return `CharacterAppearance(is_main=True)`, replay `_publish_characters`, and assert the stored flag; current result is `AssertionError: assert False is True`.

src/novelvideo/structured_builders.py
```diff
@@ -198,6 +198,72 @@ async def finish(status_error: tuple[str, str]) -> None:
...
+    Field by field rather than wholesale: a character may have been given a
+    role by hand and left without a face, and overwriting the first to supply
+    the second would lose an edit the user made deliberately.
+    """
+    repaired = 0
+    for name, appearance in appearances.items():
+        if name in added:
+            continue
+        existing = store.get_character(name)
+        if existing is None or str(existing.face_prompt or "").strip():
```
Handanhhhy: [P2] Do not skip the whole field-by-field repair just because `face_prompt` is already populated. A user may have written only the face while `role`, `body_type`, and voice remain empty; this early `continue` prevents all of those empty fields from ever being enriched, despite preserving the user-written face being possible by simply omitting it from `updates`. Minimal red test: store a character with only `face_prompt`, replay an appearance containing role/body/age, and assert the empty fields are filled; current result leaves `role == ""`.

Handanhhhy (changes requested):
Requesting changes for two reproducible gaps in the existing-character repair path. Both findings were verified against this exact head with minimal regression tests; the rest of tests/test_structured_builders.py passes (88 passed), as do ruff and git diff --check.

lywaterman:
Both findings hold. I reproduced each on the reviewed head before changing anything — the narrator test and the hand-written-face test both fail there. Fixed in `c1ce0083`.

**[P1] `is_main` on the repair path.** You are right, and the consequence you named is exact: `seedance2_i2v/voice_clone.py:292` refuses with `未找到解说主角（is_main=True）`. `_apply_appearance` set the flag on new rows and the repair path did not, so any project built before this stage existed kept every character at False permanently. A field the stage decides but never writes back is a hole, not a conservative default.

It is written now, with one condition. `is_main` has no empty state and nothing binds to it, so — unlike the age band, where an unbound voice is what marks a band nobody chose — the only signal that the question is still open is that no character holds the flag. `_enforce_single_main` already guarantees at most one narrator inside a single appearance result, and `get_all_characters` runs after the new rows are inserted, so a narrator claimed by a newly *added* character is visible to the repair loop as well. What is left is a narrator somebody marked by hand, and that one wins. There is a test for that case too.

**[P2] The early `continue`.** Also right, and your framing is the part I got wrong: preserving a user-written face by omitting it from `updates` was available all along, so gating the entire row on it was never necessary. The face was standing in for "this character has been through the stage", which is a bad proxy — it punishes someone for the one field they filled in.

Every field is now guarded on its own emptiness and on nothing else. A character with nothing missing produces an empty `updates` and is skipped before the write, so a settled cast still costs no writes; there is a test asserting `update_character` is never called in that case.

Four new tests, two of them red on the reviewed head. `ruff` clean, 3842 passed, 16 skipped.

Handanhhhy (approved):

Handanhhhy (approved):

PR #372, opened 2026-08-24
fix: stop materializing the graph preview on read

lywaterman:
Follow-on from the structured_v1 review. Not structured-specific — this is a legacy read path.

## What it did

`graph-preview.json` is written **during import**, before `novel.txt` — which is the public "import succeeded" marker — so any project that imported successfully has one. Projects predating the sidecar do not, and `GET /ingest/graph` materialized theirs on demand.

That backfill was a **viewer-role GET opening Ladybug from an API worker**, holding the project graph lock, from inside a request.

## Why that is bad

The lock is *shared* for reads, so readers do not queue behind each other. But a rebuild takes it **exclusively**, and the acquire loop has no deadline:

```python
while True:
    try:
        fcntl.flock(handle.fileno(), operation | fcntl.LOCK_NB)
        return handle
    except OSError:
        await asyncio.sleep(0.05)
```

So opening the import page during a rebuild hangs for the length of the rebuild — measured at **87s** for an import — and `asyncio.shield` keeps the request alive after the client has disconnected.

And **the wait bought nothing**. On acquiring the lock, the first thing it did was re-check `novel.txt`, find it removed by the rebuild, and return the empty preview. The guard the normal path applies *before* touching anything ("Do not race the active/failed import … or open its embedded graph database from an API worker") had been moved to *after* the lock in this branch.

## The change

A project without a sidecar gets an empty preview.

Nothing downstream reads this file — `load_graph_preview` has exactly one caller, this route — and the client already renders nothing for an empty graph:

```tsx
{knowledgeGraph.data?.data.nodes.length ? <KnowledgeGraphVisualization … /> : null}
```

So the cost is a missing picture on the import page for projects that predate the sidecar *and* have not had that page opened since (opening it once was what backfilled them). Re-importing produces it. No data is lost, because the preview is not data — it is a bounded 48-node snapshot for display.

Removes the only place an API worker opened Ladybug for visualization, and the 503 path that existed to cover it. Net −72/+55, and `asyncio`, `HTTPException` and `make_cognee_store_for_context` are no longer needed by the module.

## Tests

- a missing sidecar reads as empty, is **still** missing afterwards (the read is a read), and the route cannot construct a Cognee store at all
- a written sidecar is served from the file
- the existing "no success marker ⇒ empty" contract keeps its meaning; it no longer needs a monkeypatch to prove Cognee is not opened, since there is no longer a path that could

`test_m06_legacy_graph_is_materialized_once` is removed — it asserted the behaviour being deleted.

`ruff` clean. Full suite **3843 passed**, 16 skipped.

### DIFF
--- a/src/novelvideo/api/routes/ingest.py
+++ b/src/novelvideo/api/routes/ingest.py
@@ -1,6 +1,5 @@
 """小说上传 & 导入端点。"""
 
-import asyncio
 import logging
 import os
 import secrets
@@ -9,7 +8,7 @@ import stat
 from pathlib import Path
 from typing import Annotated
 
-from fastapi import APIRouter, Depends, File, Form, HTTPException, UploadFile
+from fastapi import APIRouter, Depends, File, Form, UploadFile
 
 from novelvideo.api.auth import get_api_user, require_scope
 from novelvideo.api.chapter_preview import (
@@ -17,9 +16,8 @@ from novelvideo.api.chapter_preview import (
     count_billable_novel_chars,
     load_novel_text,
 )
-from novelvideo.api.deps import make_cognee_store_for_context, resolve_project_scope
+from novelvideo.api.deps import resolve_project_scope
 from novelvideo.api.schemas import IngestStart
-from novelvideo.cognee.ladybug_access import ladybug_graph_access
 from novelvideo.knowledge_pipeline import is_structured_pipeline
 from novelvideo.graph_preview import (
     empty_graph_preview,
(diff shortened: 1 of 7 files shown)


Handanhhhy (approved):

Handanhhhy (changes requested):
建议修改后再合并，当前有三个问题：

1. [P1] 新角色可能绕过已有主角保护。新候选会先带着 is_main=true 经 add_characters_atomic 写入，之后 _repair_missing_appearances 才读取已有主角；若库中原本已有手工主角，本轮新增角色又被模型标记为主角，最终会留下两个 is_main=true。现有测试预先创建了第二个角色，没有覆盖“已有主角 + 本轮新增主角”。建议在构造或插入候选前读取已有主角并压掉新候选的 is_main，同时补该分支测试。

2. [P2] is_main 是整份角色表的关系属性，却按单角色缓存。缓存 payload 包含 is_main，但 key 不包含完整 cast 或角色顺序。新增、删除或重新裁决其他角色时，旧角色仍可能命中历史 is_main=true；后面的单主角裁剪只按当前顺序保留第一个，并不会重新判断真正的解说主角。建议将 is_main 从单角色外观缓存拆出，或把完整有序 cast 指纹加入缓存合同。

3. [P2] 路由把 sidecar 损坏或读取失败伪装成合法空图。load_graph_preview 对 FileNotFoundError、OSError、JSONDecodeError 和结构错误都返回 None，新路由再统一返回 200 + empty graph。当前项目的 graph-preview.json 若截断、损坏或暂时不可读，用户和监控都会看到正常空图且不会恢复。建议只对真正不存在的旧项目返回空图；已存在但损坏或不可读应记录并返回可观测失败。

另外，PR 标题和正文只描述图谱读取修复，但当前 diff 还包含 structured_v1 角色外观、主角和语音功能的大批改动。如果这是堆叠 PR，请调整 base；否则建议拆分，避免事故修复与新功能绑定发布和回滚。

移除 GET 请求内 Ladybug 回填的方向本身正确。

PR #375, opened 2026-08-24
fix: settle the narrator once, ahead of every write, and stop caching it

### LINKED ISSUE #373: bug: structured_v1 角色补全可能产生多个解说主角并复用过期 is_main
## 背景

由已合并 PR #372 引入或暴露。合并提交：952731dbc0b45301db0245e8fe79bf57a7cde09f。

## 问题一：新增角色绕过已有主角保护

_publish_characters 先通过 _apply_appearance 把模型返回的 is_main 写入新候选，再调用 add_characters_atomic。_repair_missing_appearances 在插入完成后才读取已有主角。

当数据库中已有一个 is_main=true 的角色，而本轮新发现的角色也被模型标记为主角时，新角色会先以 is_main=true 落库。后续 repair 只阻止再次设置 true，不会清除已经写入的值，因此最终可能留下两个主角。

现有 test_a_narrator_marked_by_hand_is_not_joined_by_a_second_one 预先创建了第二个角色，只覆盖 repair 既有行的分支，没有覆盖“已有主角 + 本轮新增主角”。

## 问题二：is_main 使用了错误的缓存边界

is_main 是依赖完整角色集合和顺序的关系属性，但 character_appearance_cache_key 只包含单个角色的数据和 synopsis；缓存 payload 却包含 is_main。

新增、删除或重新裁决其他角色时，旧角色仍可能命中历史 is_main=true。之后的 _enforce_single_main 只按当前顺序保留第一个，不会重新判断真正的解说主角；已经写入数据库的旧主角也不会被清除。

## 期望

- 写入新候选前读取并执行统一的单主角策略，保证提交后最多一个 is_main=true。
- 明确手工主角与模型生成主角的优先级。
- 将 is_main 从单角色外观缓存中拆出，或将完整有序 cast 指纹纳入缓存合同。
- 增加以下回归测试：
  - 已有手工主角 + 本轮新增且被模型标记为主角的角色；
  - cast 变化但旧角色自身输入不变时，不复用过期 is_main；
  - 任意发布或重建结束后，数据库中 is_main=true 的角色数不超过 1。

## 关联

- PR: #372
- Review: https://github.com/dramaclaw/dramaclaw/pull/372#pullrequestreview-5006413977

lywaterman:
fix: settle the narrator once, ahead of every write, and stop caching it

Closes #373. Both reports reproduce on staging exactly as written; the tests
for them were red before anything changed:

    assert ['郑玉琴', '林某'] == ['郑玉琴']
    assert ['郑家悦'] == ['林某']

(The issue attributes this to #372. It came from #367 — `git log -S
_repair_missing_appearances` lands on 145fd93c.)

## A new character could walk past the guard

`_apply_appearance` writes the flag onto new candidates and
`add_characters_atomic` inserts them; `_repair_missing_appearances` only
looks for an existing narrator afterwards, and only ever refuses to *add* a
second one. So a build that discovers a new character the model nominates
inserts that row with the flag already set, next to a narrator somebody had
marked by hand, and nothing downstream clears either. Two narrators, no way
back except by hand.

The guard was in the wrong place, not merely incomplete. It is now answered
once, before the first row is written: if anybody already holds the flag,
nobody in this result may claim it. The build seeds a narrator when the
project has none and otherwise keeps its hands off — the only shape that
cannot fight a decision made in the character workbench. The repair path
keeps no guard of its own; by the time it runs the flag is either the one
nomination this build may publish, or it is False.

## The nomination was cached under a key that cannot invalidate it

Every other field the stage produces belongs to one character and is keyed on
that character's own inputs, so a stored answer stays true for as long as
those inputs do. `is_main` ranks one character against the whole cast, and
the key cannot see the cast. Adding, removing or re-adjudicating somebody
else leaves the key identical and replays a nomination the new cast has
overtaken — and `_enforce_single_main` keeps the stale one, because it
settles ties rather than re-deciding.

It is no longer stored. A character served from the cache abstains, and only
characters actually asked this run can claim it. `appearance_from_cache_payload`
also forces the field to False whatever a row happens to carry, and the cache
version goes to 2 so rows written under the old contract are retired rather
than reinterpreted.

The consequence is deliberate: a rebuild where every character is a cache hit
nominates nobody, and the narrator already in the database stays. That is the
same answer as re-asking, without paying for it.

## Tests

Four. Three fail on staging: a newly discovered character cannot become a
second narrator, a stale nomination is not replayed, and a nomination stored
under the old contract is not read back. The fourth guards the other
direction — refusing to fight an existing choice must not mean never making
one, so a project with no narrator still gets seeded.

ruff clean. 3854 passed, 16 skipped.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/src/novelvideo/structured_builders.py
+++ b/src/novelvideo/structured_builders.py
@@ -214,6 +214,28 @@ def _source_synopsis(store: Any) -> str:
         return ""
 
 
+def _settle_narrator(store: Any, appearances: dict) -> None:
+    """Never publish a second narrator, and never demote the one already chosen.
+
+    ``_enforce_single_main`` bounds a single appearance result to one
+    nomination, but it cannot see the database.  A build that discovers a new
+    character the model nominates would insert that row with the flag already
+    set, and the repair path — which only ever runs afterwards, and only ever
+    refuses to *add* a second — has nothing left to prevent. The project ends
+    up with two narrators and no way back except by hand.
+
+    So the question is answered here, once, ahead of every write: if anybody
+    already holds the flag, nobody in this result may claim it. The build seeds
+    the narrator when the project has none and otherwise keeps its hands off,
+    which is the only shape that cannot fight a decision somebody made in the
+    character workbench.
+    """
+    if not any(item.is_main for item in store.get_all_characters()):
+        return
+    for appearance in appearances.values():
+        appearance.is_main = False
+
+
 def _apply_appearance(character: Any, appearance: Any, voice_for: Callable) -> None:
     """Write an appearance answer onto a character about to be created."""
     if appearance is None:
@@ -240,14 +262,6 @@ async def _repair_missing_appearances(
     one field they filled locked the others out. A rebuild adds what is absent
     and overwrites nothing.
     """
-    # ``_apply_appearance`` has already written the newly inserted rows, so a
-    # narrator claimed by a new character is visible here and is not claimed
-    # twice. Only one entry in ``appearances`` can be the narrator to begin
-    # with — ``_enforce_single_main`` settles that — so the only conflict left
-    # is with a narrator somebody marked by hand.
-    narrator = next(
-        (item.name for item in store.get_all_characters() if item.is_main), None
-    )
     repaired = 0
     for name, appearance in appearances.items():
         if name in added:
@@ -271,13 +285,13 @@ async def _repair_missing_appearances(
             updates["fish_voice_id"] = voice_for(
                 appearance.age_group, existing.gender
             )
-        # is_main has no empty state either, and nothing binds to it, so an
-        # unclaimed narrator is the only signal that nobody has decided. Left
-        # unwritten, a project built before this stage keeps every character at
-        # False and first-person narration fails with 未找到解说主角.
-        if appearance.is_main and not existing.is_main and narrator is None:
+        # Whether this may be claimed at all was settled before the first row
+        # was written; by here the flag is either the one nomination this build
+        # is allowed to publish, or it is False. Left unwritten, a project built
+        # before this stage keeps every character at False and first-person
+        # narration fails with 未找到解说主角.
+        if appearance.is_main and not existing.is_main:
             updates["is_main"] = True
-            narrator = name
         if not updates:
             continue
         await store.update_character(name, **updates)
@@ -309,6 +323,8 @@ async def _publish_characters(
         cache=StoreAnalysisItemCache(store),
         on_log=log,
     )
+
+    _settle_narrator(store, appearances)
 
     report(0.8, "发布角色...")
     candidates = []
--- a/src/novelvideo/structured_extraction.py
+++ b/src/novelvideo/structured_extraction.py
@@ -999,8 +999,7 @@ def _create_character_appearance_agent(agent: Any = None):
 # turned into stored fields changes.  It is part of every cache key, so a bump
 # retires stored results rather than mixing two contracts.
 #
-# 2: ``is_main`` left the payload.  Rows written under 1 carry it, and this
-# bump is what stops them from being read back.
+# 2: ``is_main`` left the payload for the cast-level key below.
 CHARACTER_APPEARANCE_CACHE_VERSION = 2
 
 CHARACTER_APPEARANCE_CACHE_TYPE = "character_appearance"
@@ -1112,15 +1111,77 @@ def appearance_from_cache_payload(payload: str) -> CharacterAppearance | None:
     appearance = CharacterAppearance(
         name=str(data.get("name") or ""),
         role=str(data.get("role") or ""),
-        # Never read back, whatever a row written under an older contract
-        # happens to carry: a replayed nomination is exactly the failure the
-        # payload docstring describes.
+        # Stated rather than left to a missing key: a nomination never comes
+        # from a per-character row, for the reason the payload docstring gives.
         is_main=False,
         age_group=normalize_age_group(data.get("age_group", "")),
         body_type=str(data.get("body_type") or ""),
         face_prompt=str(data.get("face_prompt") or ""),
     )
     return appearance if is_usable_appearance(appearance) else None
+
+
+CHARACTER_NARRATOR_CACHE_TYPE = "character_narrator"
+
+
+def narrator_cache_key(order: list[str]) -> str:
+    """Hash the whole cast, because that is what the answer depends on.
+
+    Deliberately not the per-character key.  Who narrates is a ranking over the
+    cast, so it is stored once per cast and retired the moment the cast changes
+    — which is exactly the invalidation a per-character key could not express.
+    """
+    payload = {"v": CHARACTER_APPEARANCE_CACHE_VERSION, "cast": list(order)}
+    canonical = json.dumps(payload, ensure_ascii=False, sort_keys=True)
+    return hashlib.sha256(canonical.encode("utf-8")).hexdigest()
+
+
+async def _settle_narrator_nomination(
+    appearances: dict[str, CharacterAppearance],
+    order: list[str],
+    cache: Any,
+) -> None:
+    """Reduce to one nomination, then remember it against this exact cast.
+
+    Storing it is what makes the stage recoverable.  Nominations are not kept
+    in the per-character rows, so a build interrupted after those rows were
+    written but before the characters were published would replay every
+    character as an abstention on retry and leave the project with no narrator
+    at all — the very failure this field exists to prevent.  The cast-level
+    entry survives that gap; a changed cast retires it and the question is put
+    to the model again.
+    """
+    _enforce_single_main(appearances, order)
+    if cache is None or not order:
+        return
+    key = narrator_cache_key(order)
+    chosen = ""
+    for name in order:
+        appearance = appearances.get(name)
+        if appearance is not None and appearance.is_main:
+            chosen = name
+            break
+    if chosen:
+        await _maybe_await(
+            cache.save(
+                CHARACTER_NARRATOR_CACHE_TYPE,
+                {key: json.dumps({"name": chosen}, ensure_ascii=False)},
+            )
+        )
+        return
+    stored = await _maybe_await(cache.get(CHARACTER_NARRATOR_CACHE_TYPE, [key]))
+    payload = (stored or {}).get(key)
+    if not payload:
+        return
+    try:
+        remembered = str((json.loads(payload) or {}).get("name") or "")
+    except (TypeError, ValueError):
+        return
+    # Only a name still in this cast, and still answered for: a nomination for
+    # somebody adjudication has since merged away must not resurrect them.
+    appearance = appearances.get(remembered)
+    if appearance is not None:
+        appearance.is_main = True
 
 
 def _enforce_single_main(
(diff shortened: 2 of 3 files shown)


Handanhhhy:
这版已经修复了 #373 描述的两条正常主路径，不过还有两个边界需要补上：

1. **已有的双主角不会在构建中收敛。** `_settle_narrator()` 只判断数据库里是否“至少有一个” `is_main=true`，随后清空本轮提名；如果旧版本已经留下两个主角，发布/重建后仍然是两个。这不满足 #373 中“任意发布或重建结束后，数据库中 `is_main=true` 的角色数不超过 1”的验收条件。虽然角色列表 GET 目前会修复历史重复值，但构建正确性不应依赖用户先访问该接口。建议在 settlement 阶段确定并保留一个现有主角，同时清除其余重复值，并增加历史双主角的回归测试。

2. **缓存写入后、数据库发布前中断，重试可能永远没有主角。** `enrich_character_appearances()` 会按 batch 先保存缓存，之后 `_publish_characters()` 才写角色。现在缓存不再保存 `is_main`，所以若任务在两者之间中断，重试时所有角色都命中 v2 缓存并以 `is_main=false` 恢复；数据库此时也没有主角，`_settle_narrator()` 不会补出提名，最终项目保持零主角，可能继续触发“未找到解说主角”。建议补一个“缓存已落盘、角色尚未发布”的全缓存重试测试，并为无现有主角且本轮无人提名的情况提供可恢复的 cast 级裁决。

除此之外，“已有手工主角 + 新增模型主角”和 cast 变化后的过期单角色缓存这两条修复方向是正确的。

lywaterman:
Thanks — point 2 is a hole this PR opened, and it is fixed in `81b0aa1d`. Point 1 I am declining, with a reason below rather than a disagreement about the mechanics; your reading of the code is correct in both cases.

**[2] Cache landed, characters did not.** You traced it exactly. Taking `is_main` out of the per-character rows removed the nomination from the only place that survived a crash, so a build killed between the batch saves and `_publish_characters` replays every character as an abstention, finds no narrator in the database either, and settles on none — permanently, until something else invalidates the appearance cache.

The fix is this PR's own argument followed through. `is_main` ranks one character against the whole cast, so it belongs at cast level rather than per character: a nomination is stored under a key hashed from the whole ordered cast; when nobody nominates because every character came from the cache, that decision is read back; when the cast changes the key changes, nothing is found, and the model is asked again. That is precisely the invalidation boundary a per-character key could not express, so recovering the nomination this way cannot bring back the stale one this PR removed. A remembered name that adjudication has since merged away is ignored rather than resurrected.

**[1] Converging pre-existing duplicates.** structured_v1 is not released. No project outside a working copy can hold two narrators, and a working copy can rebuild. Repairing data that only an unreleased revision could have written is compatibility machinery for a population of zero, so the acceptance criterion in #373 — at most one `is_main=true` after any publish or rebuild — is met going forward by settling the flag ahead of every write, and there is nothing behind us to converge.

I did implement it before deciding that, and the attempt is worth recording as a price signal rather than as an argument. To avoid two copies of the rule I routed the character list endpoint's repair through a shared store helper — real surface area on a read path serving both tracks. The helper landed between `@_auto_lease_public_async_methods` and `class SQLiteStore`, so the decorator wrapped the function instead of the class and every store method silently lost its lease; `test_sqlite_store_close_waits_for_inflight_operation` caught it. Nothing there is inherent to the idea, but it is not a cost I want to carry for data that does not exist.

Two tests added, both red on the previous head. Three removed: two for the convergence no longer attempted, one for reading a nomination out of an older payload, which only mattered if unreleased rows had to be honoured. `ruff` clean, 3855 passed, 16 skipped.

Handanhhhy:
Thanks for the update. Per the clarified scope, I am withdrawing point 1 about converging data written by an unreleased revision; that is not a blocker for this PR.

The cast-level cache is the right boundary for the nomination, but I think the crash hole is narrowed rather than closed:

1. **The per-character rows and the cast nomination are still separate writes.** Each batch saves its `character_appearance` rows inside `run_batch()`; only after all batches return does `_settle_narrator_nomination()` save `character_narrator`. If the process dies after every per-character row has landed but before that final save, the retry gets no pending characters, finds no nomination in the restored appearances, finds no cast-level entry, and returns the same all-abstention result. The new retry test completes the first `enrich_character_appearances()` call, so both kinds of cache have already landed; it does not exercise this remaining interruption point. A regression test can pre-populate only the per-character entries (with no `character_narrator` row) and then run the all-hit retry.

2. **The cast key fingerprints names/order only, not the inputs being adjudicated.** `narrator_cache_key(order)` remains unchanged when a character's description/evidence or the synopsis changes while the ordered names stay the same. The per-character keys can invalidate and cause a fresh model answer, but if that answer contains no nomination, the code falls back to the old cast decision under the unchanged names-only key. Using the ordered per-character input/cache fingerprints (including the shared synopsis contract), rather than names alone, would make this a full ordered-cast cache boundary.

So the original #373 paths are fixed, and pre-existing duplicate repair is out of scope; these two points concern only state produced and retried by the current implementation.

lywaterman:
Both hold. Fixed in `3e8ae5bf`.

**[1] The window moved rather than closed.** Correct, and your sequencing is exact — rows inside `run_batch`, nomination once after every batch returns.

I took the fix in a different direction than the regression test you proposed, because that test constructs a state the code should not be able to reach. Rather than recovering from "all rows present, no nomination", the two writes are now ordered so that state cannot be produced: a batch stores its own nomination **first**, then its rows. The rows are what make a character look answered to the next run, so putting them last means a crash between the two leaves the batch pending, and the retry asks the question again and nominates again.

The final write stays, and its job is now only determinism. Batches run concurrently, so the one that stores first is not necessarily the one earliest in cast order; the settled winner overwrites whatever they raced to.

Tested as an invariant rather than as a recovery: one test asserts the nomination write precedes the rows write, and another kills the rows write and asserts the character is replayable afterwards instead of silently answered.

**[2] The key did not cover its own dependency.** Also correct, and this one was simply an oversight on my part — I had just argued in this PR that the nomination depends on the whole cast, then keyed it on the one part of the cast that survives every edit.

It now hashes the per-character cache keys in cast order. Each of those already covers that character's whole input including the shared synopsis, so the ordered list is the full dependency and the entry retires whenever any part of it does. The end-to-end test is the load-bearing one: same cast, same order, edited description, fresh answer nominating nobody — the old decision must not stand in, and on the previous head it did.

Four tests, two red on the previous head. `ruff` clean, 3859 passed, 16 skipped.

Handanhhhy:
The two previous points are fixed in `3e8ae5bf`. I found one remaining crash-consistency edge when more than one batch nominates a character:

**The pre-row writes race on one cast key, so a crash before the final settlement can recover the wrong nominee.**

Each `run_batch()` writes its own `_nominee(produced, order)` to the same `narrator_key`, then writes that batch's appearance rows. With multiple batches, more than one batch can legitimately return `is_main=true`; `_enforce_single_main()` exists specifically to settle those answers in cast order. On the successful path, the final `_settle_narrator_nomination()` overwrites the raced value with that deterministic winner.

There is still a reachable interrupted state, though:

1. An earlier cast batch nominates A and writes A plus its rows.
2. A later batch nominates B and overwrites the same cast key, then writes its rows.
3. The process dies after all rows are present but before the final settlement stores A.
4. On retry every appearance is a cache hit and therefore abstains; the cast cache restores B, even though the uninterrupted result would have selected A.

So ordering nomination before rows guarantees that *a* nomination survives, but not that the globally settled nomination survives.

A focused regression test could use two batches that both nominate, run them deterministically so the later batch writes last, fail the final `CHARACTER_NARRATOR_CACHE_TYPE` save after all appearance rows have landed, then retry from the cache. The expected recovered narrator should match the uninterrupted cast-order winner.

Possible fixes include persisting recoverable per-batch decisions under separate keys and reducing them on retry, or atomically coupling the final cast decision with the state that makes every batch appear complete.

lywaterman:
Holds, and it is the sharper version of the previous point: ordering guaranteed that *a* nomination survives, not that the settled one does. Fixed in `60adb59e`.

My first attempt at your regression test was wrong in a way worth naming, because it made the bug look absent. I let the first `enrich_character_appearances()` call complete, so the final settlement ran and stored the correct winner — the test passed against the broken head. The crash has to be injected at the point you named, not merely somewhere near it: every appearance row on disk, the settlement not yet stored. With the injector in the right place it reproduces exactly as you described:

    assert ['林某'] == ['郑家悦']

**The fix removes the race instead of correcting it afterwards, and is smaller than what it replaces.** Each nominee now gets its own key, derived from the cast fingerprint and that character's own fingerprint. Nothing is overwritten, so the batches need no ordering among themselves, and the reduce moves to the read: a run with no nomination of its own reads the votes back and takes the earliest in cast order — the same rule `_enforce_single_main` applies in memory, so an interrupted run and an uninterrupted one reach the same answer by construction rather than by repair.

The final settlement write is gone with it. Its only job was undoing the race, and there is no longer a race to undo. Of your two suggested directions this is the first one; I did not need the atomic coupling, because separate keys make the write order irrelevant rather than merely safe.

The cast fingerprint stays inside every vote key, so a changed cast retires the votes along with the per-character rows they were decided from — the boundary from the previous round is unaffected.

Two tests, both red on the previous head. `ruff` clean, 3861 passed, 16 skipped.

Handanhhhy:
One scope clarification for the reviewer:

The remaining partial-retry scenario is not compatibility with an older deployed version or old production data. It can be created entirely by the first run of this new implementation: an earlier character in cast order records its vote and appearance row, the worker stops before a later batch completes, and the retry combines that cached character with a fresh later nominee. Because `_settle_narrator_nomination()` returns as soon as the current in-memory appearances contain any nominee, it does not read the earlier persisted vote, so the retry can choose a different narrator from an uninterrupted run.

That said, this feature has not been deployed, and the requested scope is the direct #373 fix rather than guaranteeing identical narrator selection across partial task retries. I am therefore treating this as **non-blocking for PR #375** and leaving it to the maintainers/reviewer to decide whether the resumability contract expressed by the cache comments and tests warrants handling it here or separately.

lywaterman:
Thanks for separating the scope question from the finding — you are right on both counts, and I am taking it rather than deferring it. Fixed in `c2b3e22a`.

Three reasons it is worth doing here rather than separately. It is reachable from an ordinary partial retry, not a crash window, so it is the resume path working as designed rather than a rare interleaving. The comments and tests in this PR already claim the resumability contract it breaks, so leaving it would mean shipping documentation that is wrong. And the fix removes an asymmetry instead of adding a mechanism.

Your diagnosis is exact: the early return preferred the in-memory answer, and a character replayed from cache abstains in memory however loudly an earlier attempt nominated it — which is the direct consequence of keeping nominations out of the per-character rows.

    assert ['林某'] == ['郑家悦']

Since batches write their votes ahead of the rows that make them look finished, the votes are a superset of what is in memory by the time settlement runs, so there was never a reason to prefer the smaller set. The early return is gone and the reduce always runs over the votes in cast order — the same rule `_enforce_single_main` applies — which makes the answer independent of where a build was interrupted as well as of the order its batches finished in. Votes are matched only against characters this run actually has an appearance for, so one belonging to a character adjudication has since merged away, or whose own row failed this time, cannot bring them back.

One test, red on the previous head: a first attempt answers the earlier character and stops before the later one; the retry replays it silently and answers the later one, who nominates himself; cast order still decides.

`ruff` clean, 3862 passed, 16 skipped.

Handanhhhy (approved):

PR #410, opened 2026-08-26
fix(brainclaw): keep one project-scoped session per canvas

lywaterman:
## Summary
- derive canvas trajectory identity from project + canvas + Agent profile
- prevent canvas IDs from merging evidence across projects and preserve future multi-session separation
- keep all multi-session creation/history implementations intact
- hide both creation actions and the history-session UI behind one explicitly disabled product flag

## Verification
- 50 focused BrainClaw/Control Context tests passed on the exact PR head
- same canvas ID across projects yields distinct trajectories
- different Agent profiles on one project/canvas yield distinct trajectories
- 12 focused frontend tests passed
- TypeScript build check and full frontend production build passed

### DIFF
--- a/frontend/src/__tests__/features/freezone/canvas-agent-single-session-contract.test.ts
+++ b/frontend/src/__tests__/features/freezone/canvas-agent-single-session-contract.test.ts
@@ -1,0 +1,26 @@
+// SPDX-License-Identifier: Elastic-2.0
+// Copyright (c) 2026 ClaymoreLab
+import { readFileSync } from "node:fs";
+import { resolve } from "node:path";
+
+import { describe, expect, it } from "vitest";
+
+describe("Freezone canvas Agent session UI", () => {
+  it("keeps multi-session code but hides both creation actions", () => {
+    const shell = readFileSync(
+      resolve(process.cwd(), "src/features/freezone/FreezoneShell.tsx"),
+      "utf8",
+    );
+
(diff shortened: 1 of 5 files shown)


vkiki:
I found two runtime concerns that should be addressed before merge:

1. When FREEZONE_MULTI_AGENT_SESSION_UI_ENABLED is false, agentState.agents is still mapped into a SuperChatPanel for every historical Agent. An inactive panel can initialize with busy=true from a cached pending turn, report that through onConnectionStateChange, enter busyAgentIds, and then satisfy active || busy, reconnecting a hidden legacy session. This can leave hidden WebSockets or approvals running and means the single-session mode is not fully enforced. Please render only the canonical main panel while the flag is disabled, and restore the full mapping only when multi-session UI is enabled.

2. Existing users whose active conversation is an agent-* session lose all UI access after this change: initialization forces main, the URL agent override is ignored, and the history panel is hidden. The data remains stored but becomes unreachable. Please either migrate the previously active session to the canonical session, retain a read-only history/recovery entry, or explicitly document and provide a recovery path.

The new frontend test currently checks source strings, so it does not cover either state transition. A behavioral test with a stored non-main active Agent and a cached pending turn would catch both regressions.

The backend trajectory identity change to project + canvas + Agent profile looks correct and addresses cross-project canvas-ID collisions.

PR #411, opened 2026-08-27
refactor(codex): keep ephemeral files in runtime

lywaterman:
## Summary

- keep durable Codex rollouts and SQLite under `state/.codex-app-server`
- keep node-local operating-system IPC under `/tmp/claymore-<uid>/<home-node-id>/`
- keep App Server logs under `NOVELVIDEO_RUNTIME_DIR/codex/logs`
- move per-turn gateway token files out of project state into `NOVELVIDEO_RUNTIME_DIR/codex/turn_tokens`
- enforce private `0700` directories and `0600` log/token files
- do not migrate legacy temporary files during the development phase

## Storage contract

- `output`: OSS-backed persistent project working set; unchanged
- `state`: node-local consistency state that cannot safely live on OSS; Codex home remains here
- `runtime`: disposable DramaClaw application logs and per-turn credentials
- `/tmp/claymore-<uid>/<home-node-id>/`: disposable operating-system process-control files (`node-runtime.lock`, `app-server.signature`, `app-server.sock`; future PID/token files belong here too)

The fixed `/tmp` control path is intentional: Unix-domain sockets have a short platform path limit, especially on macOS, so they must not inherit an arbitrarily long application runtime path.

## Verification

- `uv run ruff check src/novelvideo/chat/codex_app_server.py src/novelvideo/chat/service.py tests/test_codex_app_server.py tests/test_chat_service_user_agent_scope.py`
- `uv run pytest -q tests/test_docker_persistence_config.py tests/test_sandbox_roots.py tests/test_chat_service_user_agent_scope.py tests/test_codex_app_server.py` — 198 passed
- wheel release artifact test passed with network-enabled isolated build
- full backend run reached 4688 passed / 16 skipped; its only remaining code-independent failure was the staging license inventory omission for `canvas-agent-single-session-contract.test.ts`, which PR #410 now fixes
- `git diff --check`

### DIFF
--- a/src/novelvideo/chat/codex_app_server.py
+++ b/src/novelvideo/chat/codex_app_server.py
@@ -72,13 +72,14 @@ def _deny_unexpected_approval(method: str, _params: dict | None) -> dict:
 
 
 def _control_dir() -> Path:
-    """Return a short, private directory for the home-node control socket."""
+    """Return the short, private system root for node-local IPC files."""
 
     # macOS limits Unix-domain socket paths to roughly 104 bytes. CODEX_HOME is
-    # intentionally below the configured state root and can easily exceed that
-    # in a local checkout, so the transport endpoint cannot live below it.
+    # intentionally below the configured state root and can easily exceed that.
+    # These files are operating-system IPC state, not DramaClaw runtime data, so
+    # keep them under a fixed short path instead of following TMPDIR/runtime.
     uid = os.getuid() if hasattr(os, "getuid") else os.getpid()
-    path = Path("/tmp") / f"dramaclaw-codex-{uid}"
+    path = Path("/tmp") / f"claymore-{uid}"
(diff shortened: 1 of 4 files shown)


vkiki (commented):
Thanks for the runtime-boundary refactor. The storage split and permission hardening look reasonable, but I found two items to address before merge:

1. `_app_server_log_path()` creates/chmods the log directory and file without rejecting symlinks. If a log path is pre-created as a symlink, `touch/chmod/open("ab")` may follow it and write or chmod an unintended file. Please add the same `lstat()` symlink/type checks used for the control and token directories, plus a regression test.

2. The PR description says the App Server socket, lock, and signature move under `NOVELVIDEO_RUNTIME_DIR/codex`, but the implementation keeps them under `/tmp/claymore-{uid}` (presumably for macOS Unix-socket length limits). Please clarify the contract or align the description with the implementation.

The reported CI checks look good otherwise.

PR #416, opened 2026-08-27
fix(agent): stabilize workflow planning and approvals

vkiki:
## 目的

统一 Hermes、Codex App Server 与第三方 Agent 使用的 Workflow Skill/MCP 契约，修复参数确认、画布批量写入、错误回显和工作流 Schema 漂移问题。

## 主要改动

- 提取公共 WorkflowPlan/WorkflowIntent JSON Schema，并由 Hermes 与独立 MCP 复用
- 补全节点参数、连接、分组、宫格布局及 run_after_create 契约
- 修复执行标志别名优先级、空 prompt 覆盖和分组引用校验
- 完善图片/视频参数确认及审批流程，保持只创建不执行场景不触发生成
- 增加 Skill/Recipe/MCP 真实 stdio 端到端诊断程序
- 修复 social-content-campaign 内置 Skill 被 Schema 静默跳过的问题
- 同步可发布 Agent Skill 文档，并持久化聊天错误供前端展示

## 验证

- Workflow MCP/Skill/Recipe 诊断：702 checks，0 errors
- 后端全量：4724 passed，16 skipped，2 deselected
- 前端全量：3774 passed
- Ruff 与 git diff --check 通过

## 兼容性

- 不修改 CI
- 保留 Hermes 原有授权、审批和画布提交边界
- 节点模型参数继续通过实时 Schema 渐进式加载


lywaterman (changes requested):
Blocking findings:

1. **[P1] Distinct workflows on the same canvas are silently discarded as duplicates**

`activeOrQueuedWorkflowRunsByCanvas` is keyed only by `canvasId`. While workflow A is active or queued, a legitimate, different workflow B on the same canvas hits the early return in `executeQueuedNodeActions`, is not queued, and is nevertheless reported as `status: "success"` with `workflow_already_running`. The regression test only resubmits the same envelope, so it does not cover two distinct workflow targets.

Please deduplicate by a stable workflow/run identity (for example canvas + workflow instance/run id or a normalized command fingerprint), while continuing to serialize distinct workflows through the canvas queue.

2. **[P1] External agents can bypass the required generation-parameter clarification**

The generic canvas-command validator accepts `{"type":"run_workflow"}` without either `node_ids` or `scope: "canvas"`. For that shape, `_workflow_generation_target_ids()` starts from an empty set, so `_external_generation_parameter_preflight()` returns success without inspecting the frontend's current selection. The frontend then interprets the same command as “run the selected workflow”; in `auto_execute` mode this can execute generation using defaults without the mandatory clarification introduced by this PR.

I reproduced this as `shape_error=null` and `generation_preflight=null` against a selected image-generation node missing all required generation fields. Please enforce the same `node_ids or scope=canvas` rule already used by `freezone_run_workflow`, or resolve the actual selection before preflight.

3. **[P2] Portable generation inputs are copied into WorkflowPlan without validation**

The new universal generation keys bypass the Skill parameter validation and are copied directly into `resolved`. Counts are then written into node data whenever they are integers. For example, `image_count=-7` compiles successfully and produces a plan with `data.count=-7`, contradicting this PR's own WorkflowPlan JSON Schema (`minimum: 1`). Other universal model/aspect/quality values are similarly unchecked.

Please validate types, supported values, and ranges for these portable inputs in one place and add invalid-value regression tests. Downstream clamping does not fix the contract because it silently changes the value the user confirmed.

CI is also currently blocked:

- DCO fails because both commits lack `Signed-off-by`.
- `deps` fails because the license inventory is missing:
  - `scripts/check_workflow_agent_contracts.py`
  - `src/novelvideo/freezone/workflow_schema.py`
  - `tests/test_chat_error_persistence.py`

Focused verification performed:

- backend workflow/MCP/plugin/error tests: 153 passed
- changed frontend suites: 329 passed
- workflow agent contract checker: 698 checks, 0 errors, 1 warning
- `git diff --check`: passed

lywaterman (changes requested):
Re-review of `21b3eb55`:

The previous workflow deduplication and empty-target `run_workflow` findings are fixed, and the portable-input validation is improved. One blocking contract issue remains.

**[P1] `image_count` conflates workflow item count with per-node generation count**

`_UNIVERSAL_GENERATION_INPUT_KEYS` treats `image_count` as a portable execution parameter, and `_intent_item_node()` copies it into every image node as `data.count`. However, the built-in `social-content-campaign` Skill already defines `image_count` as the number of image workflow items and explicitly uses it to decide how many nodes to create.

I reproduced this with the real built-in Skill: compiling three social image items with its default `image_count=3` succeeds and produces three image nodes whose counts are `[3, 3, 3]`. In `auto_execute`, that can submit nine image outputs instead of the requested three. In manual approval, the UI only supports counts `1/2/4`, so each `count=3` is instead silently normalized to `1`. Both paths violate the confirmed input.

Please separate the concepts, for example keeping `image_count` as workflow deliverable/node count and introducing `image_variants_per_node` or `image_generation_count` for per-node outputs. Apply the same separation to video, validate per-node counts against the actual supported values, and add a regression proving that `social-content-campaign` with three image items does not write `count=3` into every node.

Focused verification:
- backend workflow/plugin/MCP/Codex/error suites: 186 passed
- chat service scope suite: 173 passed
- changed frontend suites: 330 passed
- workflow contract checker: 698 checks, 0 errors, 1 existing orphan-Recipe warning
- current CI: green, including DCO and dependency inventory

lywaterman (changes requested):
Re-review of `f5d8231c`:

The workflow-count versus per-node-variant split is now correct. The real `social-content-campaign` regression passes, variant counts are limited to `1/2/4`, and the related Skill/Codex/Hermes contracts are aligned.

One portable-input validation issue remains:

**[P2] Reject invalid values for stable generation enums before compiling the plan**

`_portable_generation_input_error()` currently validates `image_aspect_ratio`, `image_resolution`, `image_quality`, `video_aspect_ratio`, and `video_resolution` only as non-empty strings. Several of these are stable portable enums, not arbitrary model ids.

I reproduced this with the real `text-to-image-video` Skill. Inputs containing `image_aspect_ratio="banana"`, `image_resolution="8K"`, and `image_quality="ultra"` compile successfully and are copied unchanged into the image node as `aspectRatio`, `size`, and `quality`.

The frontend contract only supports image sizes `1K/2K/4K`, image quality `low/medium/high`, fixed image/video aspect-ratio sets, and the global video-resolution set `480P/720P/1080P`. Manual approval preserves the invalid strings unless the user changes them, while `auto_execute` applies the original approval directly. The resulting generation can therefore fail late or silently fall back after the user-facing settings were already considered confirmed.

Please validate the stable portable enums in the compiler. Keep model ids dynamic, and let the live schema continue narrowing model-specific subsets such as supported video resolutions and durations. Add regressions rejecting values such as `banana`, `8K`, and `ultra`.

Verification on this head:
- backend workflow/plugin/MCP/Codex/error suites: 187 passed
- chat service scope suite: 173 passed
- relevant frontend suites: 359 passed
- workflow contract checker: 698 checks, 0 errors, 1 existing orphan-Recipe warning
- CI: green

vkiki:
Fixed in `a8388e65`.

The compiler now rejects invalid values for the stable portable generation enums before building the WorkflowPlan:

- `image_aspect_ratio`: the fixed approval-card aspect-ratio set
- `image_resolution`: `1K` / `2K` / `4K`
- `image_quality`: `low` / `medium` / `high`
- `video_aspect_ratio`: the fixed approval-card video aspect-ratio set
- `video_resolution`: `480P` / `720P` / `1080P`

Model ids remain dynamic. The progressively loaded live schema can still narrow these portable values for a selected model, including model-specific resolution and duration subsets.

Added regressions for all five enum fields plus real built-in `text-to-image-video` coverage rejecting `banana`, `8K`, and `ultra`.

Verification:
- workflow plan + freezone plugin tests: 164 passed
- workflow contract checker: 702 checks, 0 errors (3 existing local warnings)
- Ruff and diff checks: passed

lywaterman (changes requested):
Re-review of `a8388e65`:

Correction to my previous review: after checking the authoritative canvas media-model catalog, asking the portable compiler to maintain a global stable enum set was incorrect. The latest commit implements that request, but the duplicated enum now diverges from the models that the canvas actually exposes.

**[P1] Validate generation capabilities against the live canvas model schema, not a parallel global enum**

`_PORTABLE_GENERATION_ENUM_OPTIONS` rejects legitimate model-specific values. Concrete examples from the current official catalog include:

- `seedream-5.0-lite` supports image resolution `3K`.
- `LingShan-NB-2` supports `1:4`, `4:1`, `1:8`, and `8:1` aspect ratios.
- Video models expose values outside the new list, including `4K`, `768P`, `2K`, and additional ratios such as `9:21`, `5:4`, and `4:5`.

The canvas already treats its scoped Media Model Catalog / `freezone_get_node_create_schema` result as the authoritative source for model ids and dependent capabilities. The Agent path needs to use that same source. Please remove the independently maintained global enum table. The portable compiler can continue checking scalar types and non-empty strings; exact model, ratio, resolution, and quality compatibility must be validated against the selected model from the live canvas schema before the protected write. If early compiler validation is required, inject the same scoped catalog rather than copying another allowlist.

The existing direction for model ids is correct: use the exact catalog id returned by the live schema rather than provider aliases or display labels.

CI note: all other current checks are green, but DCO fails because the latest commit lacks `Signed-off-by`.

Once the parallel enum is removed/replaced with the canvas-authoritative validation and DCO is fixed, this finding is resolved.

vkiki:
Addressed in `01c7152b` (supersedes `a8388e65`).

Thanks for the correction. I removed the parallel portable enum table and restored the compiler boundary to scalar/type/non-empty validation for model-dependent strings.

Capability compatibility is now checked against the project-scoped live image/video model catalog during workflow runtime preflight, before canvas commands are built/emitted. The selected catalog model is used to validate:

- image `aspectRatio`, `size`, and `quality`
- video `aspectRatio`, resolution (`quality`), `durationSec`, and `generateAudio`

The same preflight now covers draft prepare/confirm, direct `freezone_create_workflow_graph`, and the compact intent compatibility entrypoint. Model ids remain the exact ids returned by the scoped catalog. Resolution matching is case-insensitive only to preserve the catalog’s existing `2K` / `2k` compatibility; supported values still come exclusively from the selected catalog entry.

Regressions prove that compiler planning accepts legitimate model-specific values (`3K`, `1:8`, `2K`, `9:21`), live preflight accepts the matching `seedream-5.0-lite`, `LingShan-NB-2`, and `MiniMax-H3` capabilities, and rejects values outside the selected model schema.

Verification:
- workflow plan + Freezone plugin suites: 162 passed
- workflow contract checker: 702 checks, 0 errors (3 existing local warnings)
- Ruff and diff checks: passed

The amended commit includes `Signed-off-by` to resolve DCO.

lywaterman (changes requested):
Re-review of `01c7152b`: The parallel global generation enum is removed, legitimate model-specific values defer to the live catalog, and DCO is fixed. **[P1] Do not write a workflow when live model capabilities could not be verified.** `_workflow_runtime_preflight()` treats a failed image/video catalog request as a warning; callers stop only on blockers and can continue to the protected canvas write. I reproduced `{ok:false}` from the image catalog with `size=8K`, `aspectRatio=banana`, `quality=ultra`: result was `status=ready`, `blockers=[]`, warning only. The later external generation preflight checks presence, not compatibility. When media nodes need model-dependent validation, unavailable live catalog must produce a blocker such as `model_catalog_unavailable`; queue-capacity lookup may remain best-effort. Add a failure-path regression covering prepare, confirm, and direct-create. Focused local tests: 8 passed; CI green.

vkiki:
Fixed in `e034c5a2`.

`_workflow_runtime_preflight()` now fails closed when a workflow contains image/video nodes with selected models but the corresponding project-scoped live model catalog request fails. It adds a `model_catalog_unavailable` blocker instead of a warning, so callers stop before building or emitting protected canvas writes. Queue-capacity lookup remains best-effort as requested.

Added failure-path regressions proving that catalog unavailability blocks:

- workflow draft preparation
- workflow draft confirmation (including a catalog that becomes unavailable after prepare)
- direct `freezone_create_workflow_graph`

The lower-level regression also reproduces the invalid `8K` / `banana` / `ultra` payload with `{ok:false}` from the image catalog and asserts `status=blocked` with no write path.

Verification:
- workflow plan + Freezone plugin suites: 166 passed
- workflow contract checker: 702 checks, 0 errors (3 existing local warnings)
- Ruff and diff checks: passed
- commit includes `Signed-off-by`

lywaterman (approved):
Re-review of `e034c5a2`: all previously reported workflow validation and approval issues are resolved. In particular, live media-model catalog failures now block direct workflow creation, draft preparation, and draft confirmation before any protected canvas write. Local verification: 370 focused backend tests passed; workflow contract checker passed 698 checks with 0 errors (1 existing orphan-Recipe warning); git diff --check passed. Current CI is green. Approved.

PR #454, opened 2026-09-02
fix(video): restore safe provider errors for organization users

ryanhuang-duat:
## Summary

- restore concrete failure details from `content`, `error`, `error_message`, and `fail_reason` for definitive organization video failures
- redact provider URLs and credential fragments before returning messages to users
- keep `EGRESS_OPERATION_UNKNOWN` for ambiguous submit, timeout, and missing-result states
- preserve the existing `VIDEO_*` safe-code mapping

## Verification

- `uv run pytest -q tests/test_p0g4c_video_egress.py tests/test_task_credit_errors.py` (65 passed)
- `pnpm test -- src/__tests__/lib/gateway-error-classify.test.ts` (406 files / 2899 tests passed)
- `uv run ruff check src/novelvideo/shared/provider_errors.py src/novelvideo/generators/video_generator.py tests/test_p0g4c_video_egress.py`
- `uv tool run pre-commit run --files src/novelvideo/generators/video_generator.py src/novelvideo/shared/provider_errors.py tests/test_p0g4c_video_egress.py`

No database migration, frontend code change, or EE change is required.

### DIFF
--- a/src/novelvideo/generators/video_generator.py
+++ b/src/novelvideo/generators/video_generator.py
@@ -24,6 +24,8 @@ from typing import Any, Callable, Optional
 from novelvideo.shared.provider_errors import (
     classify_provider_video_task_error,
     provider_video_error_code_from_text,
+    provider_video_error_message_from_text,
+    provider_video_task_error_message,
 )
 
 import aiohttp
@@ -3647,6 +3649,7 @@ class NewApiVideoGenerator(VideoGeneratorBase):
                     safe_task_error = (
                         (
                             classify_provider_video_task_error(task)
+                            or provider_video_task_error_message(task)
                             or "EGRESS_OPERATION_UNKNOWN"
                         )
                         if organization_request
@@ -3741,7 +3744,11 @@ class NewApiVideoGenerator(VideoGeneratorBase):
                 error_message=type(exc).__name__,
             )
             safe_exception_error = (
-                _safe_video_error_code(exc, "EGRESS_OPERATION_UNKNOWN")
+                (
+                    _safe_video_error_code(exc, "")
+                    or provider_video_error_message_from_text(str(exc))
+                    or "EGRESS_OPERATION_UNKNOWN"
+                )
                 if organization_request
                 else str(exc)
             )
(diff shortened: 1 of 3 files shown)


Handanhhhy (changes requested):
[P1] Do not pass arbitrary pattern-redacted provider text across the organization boundary

`src/novelvideo/shared/provider_errors.py:180` only removes HTTP(S) URLs and a small set of named key fragments, then returns every other provider-controlled message unchanged. Both new callers pass that value into the organization-visible `VideoGenResult.error` (and the request/refund records). A failed task or submit response containing an internal non-HTTP connection URI, such as a PostgreSQL DSN with embedded credentials, therefore exposes that credential to the organization user.

I reproduced this at head `78c2290b` with:

`uv run pytest -q tests/review_test_pr454_provider_error_redaction.py`

Both the failed-task and submit-error cases fail because the complete DSN remains in the returned message. Please keep this boundary allowlisted (known `VIDEO_*` codes / explicitly normalized safe messages, otherwise `EGRESS_OPERATION_UNKNOWN`), or consume an upstream field with a defined safe contract; enumerating a few secret spellings cannot make arbitrary provider text safe.

ryanhuang-duat:
Thanks for raising this. I agree that a regex denylist cannot make arbitrary provider text universally secret-free, but that is not the intended trust boundary of this change.\n\nThis path deliberately preserves upstream diagnostic messages so organization users can troubleshoot generation failures. The redaction boundary here is limited to DramaClaw-owned values that our request path can introduce and the upstream may echo back: relay/signed HTTP(S) URLs, Authorization headers, gateway API keys, and tokens. A PostgreSQL DSN produced by the upstream is upstream-owned diagnostic data; DramaClaw does not send its own database DSN to the video gateway, so the reproduction demonstrates preservation of arbitrary upstream text, not a path that exposes a DramaClaw credential.\n\nExhaustively detecting every possible secret format in an upstream diagnostic is neither feasible nor the contract of this layer. If there is a concrete path by which a DramaClaw-owned DSN or other DramaClaw credential can enter this provider response, please point it out and I agree we should block it. Otherwise, I propose documenting this boundary explicitly and pinning it with tests that relay signed URLs and organization gateway credentials are redacted while the remaining upstream diagnostic is preserved, rather than reverting to an allowlist that removes the troubleshooting information this PR is restoring.

Handanhhhy (approved):

PR #491, opened 2026-09-06
fix(settings): let Custom mode map DC-character-builder-LLM

lywaterman:
## Why

Custom mode writes NewAPI channel `models` / `modelMapping` only for the rows listed in `frontend/src/lib/feature-models.ts`. Structured character extraction (v2.0.0) added `CHARACTER_BUILD_MODEL = DC-character-builder-LLM` on the backend but never added a frontend row, so the alias can never be mapped in Custom mode and NewAPI answers `No available channel for model DC-character-builder-LLM`. That is the first error in #490; the reporter worked around it by pointing `CHARACTER_BUILD_MODEL` at the scene-builder alias in `.env`. Official mode is unaffected (RelayClaw ships the alias).

## What

- `feature-models.ts`: `CHARACTER_BUILD` → `DC-character-builder-LLM` next to `SCENE_BUILD` in the asset-library group.
- zh / en labels for the new row.
- `configuring-models` docs (en + zh) list the alias among the logical model names.
- New test `tests/test_feature_models_cover_official_defaults.py`: every `DC-*-LLM` alias in `official_defaults` must have a frontend feature row. It fails on `main` without this change (reports `DC-character-builder-LLM`) and passes with it.

Note for the next `main → staging` sync: the same test will flag `DC-freezone-recipe-compiler-LLM`, which exists only on `staging` and also has no frontend row. Add that row on `staging` when syncing.

## Verification

- `vitest run src/__tests__/i18n/locales-json.test.ts`: 10 passed; `tsc --noEmit` clean.
- New pytest passes against this branch's `src`; `scripts/lint_banned_words.py` clean.

Addresses the mapping part of #490. The `Exceeded maximum output retries` part is separate and still under investigation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01NQrE65FQLj4endsbo3Wdsw

### DIFF
--- a/docs/en/getting-started/configuring-models.md
+++ b/docs/en/getting-started/configuring-models.md
@@ -160,7 +160,7 @@ After a profile is saved, a channel key should show “Saved” and a masked pre
 
 #### Feature models
 
-DramaClaw uses stable logical names such as `DC-scene-builder-LLM` and `DC-freezone-vision-LLM`. In Custom mode, keep those internal names and map them to real upstream models in NewAPI.
+DramaClaw uses stable logical names such as `DC-character-builder-LLM`, `DC-scene-builder-LLM` and `DC-freezone-vision-LLM`. In Custom mode, keep those internal names and map them to real upstream models in NewAPI.
 
 - Text features can use text-only models.
 - Vision features send images or video and require a suitable multimodal model.
--- a/docs/en/getting-started/configuring-models.md
+++ b/docs/en/getting-started/configuring-models.md
@@ -166,6 +166,7 @@ DramaClaw uses stable logical names such as `DC-character-builder-LLM`, `DC-scen
 - Vision features send images or video and require a suitable multimodal model.
 - Bulk fill changes drafts only; click Save Mapping afterward.
 - Hermes may use a separate model. Other `DC-*-LLM` mappings can share one upstream model or be overridden individually.
+- After an upgrade that adds a feature row (for example `DC-character-builder-LLM` in v2.0.3), the new row starts empty and is not written to NewAPI until you pick a model and click Save Mapping, or re-apply a quick profile. Until then that feature fails with `No available channel for model DC-...`.
 
 #### Embedding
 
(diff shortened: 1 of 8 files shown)


lywaterman:
Self-review follow-up (`2c97803d` + zh follow-up): existing Custom-mode installs get the new row **empty** after upgrading, and `buildChannels` skips empty rows, so `DC-character-builder-LLM` stays unmapped until the user picks a model and clicks Save Mapping (or re-applies a quick profile). Documented that in both configuring-models guides (en + zh). Also ran `ruff format` on the new test.

Handanhhhy (changes requested):
[P1] Register CHARACTER_BUILD in the product groups used by Custom settings

Location: frontend/src/lib/feature-models.ts:62

The new alias is added to FEATURE_MODEL_GROUPS, but the settings UI constructs its text rows from FEATURE_MODEL_PRODUCT_GROUPS (settings-dialog.tsx:2619). That list still omits CHARACTER_BUILD. Bulk fill also iterates those displayed groups, so it cannot populate this entry either.

For an existing Custom-mode installation, upgrading and following the documented pick-a-model / Save Mapping flow therefore still cannot configure DC-character-builder-LLM: there is no row to select, and buildChannels skips the empty entry. Character extraction continues to fail with the original missing-channel error unless the user takes another path, such as re-applying a quick profile.

Please also add productFeature("CHARACTER_BUILD") to the appropriate product group and extend the coverage test to check the groups actually rendered by settings.

Verification on b05bfcf180579fe7821d8094faaec5ad8a75ff9f: a retained local regression test imports the actual TypeScript exports, verifies that CHARACTER_BUILD is registered, and asserts that it appears in the displayed text groups. Running `node --test .review-tests/character-model-visible.test.mjs` fails with `CHARACTER_BUILD is registered but absent from the rendered and bulk-filled text feature groups`. The PR's new backend-alias coverage assertion passes, demonstrating that it does not catch this omission. GitHub CI has no failed checks; full local pytest/frontend suites were not run.

Handanhhhy (approved):

PR #500, opened 2026-09-08
feat(i18n): add Vietnamese (vi) locale

### LINKED ISSUE #489: Add Vietnamese (`vi`) locale — translation is already done, happy to contribute it
## Pre-flight

- [x] I searched existing issues and Discussions. Closest matches are #440 (English
  screenplay/canvas support) and #460 (backend runtime text migration) — neither covers
  adding a third UI locale.

---

## Problem / use case

DramaClaw currently ships `zh` and `en` only (`supportedLngs: ["zh","en"]`). Vietnam has
a large short-drama / AIGC video crowd, and a lot of them read English slowly enough that
a Chinese-or-English-only UI is a real barrier to getting through the full pipeline.

I've been running DramaClaw daily and ended up localizing it to Vietnamese for my own use.
It works, but it lives as a local patch that every auto-update wipes, so I'd much rather
have `vi` upstream than keep re-applying a mod. Offering the translation back in case it's
useful.

---

## Proposed solution

**1. Accept a `vi` locale.** The i18next + HTTP-backend setup already does all the work —
`loadPath: "/locales/{{lng}}/{{ns}}.json"` picks up a new folder with no other changes.
What's needed is small:

- add `"vi"` to `supportedLngs`
- add a third entry to the language menu in the account dropdown
- add `header.account.languageVietnamese` to the `en` and `zh` locales

**2. I can open a PR with `locales/vi/translation.json`.** Current state against v2.0.0's
`en/translation.json`: **3733 / 3999 keys (93.3%)**. Placeholders (`{{count}}`, `<0>…</0>`)
are verified to match the `en` source key-for-key.

The remaining ~7% is **deliberately left in English**, and I'd want reviewers to know why:

| Group | Why |
|---|---|
| `nodeToolbar.*Prompt`, `nodeToolbar.gridMenu.*Prompt`, `multiAngleEditor.prompt*`, `lightEditor.prompt*` / `presetPrompts.*` | These are composed into the prompt sent to the image model, not UI labels. Translating them measurably hurts output quality. Same reason `multiAngleEditor.zoomLevels/horizontalLabels/verticalLabels` and `lightEditor.directions` stay English — they're interpolated into those prompts. |
| Domain terms (Model, Prompt, Render, Timeline, Keyframe, Upscale, Seed…) | Vietnamese video/AI people use the English words; translating them reads worse. |
| Proper nouns (SuperTale, Freezone, Director World, XiaHua, XiaJing…) | Product names. |
| Screenplay format keywords in `ingest.novelFormat.*` and `aiAssistant.formatCheck.recommended` | See the note at the bottom — the parser only accepts zh / Fountain-EN keywords, so the guide explains the rules in Vietnamese while keeping the literal keywords intact. |

If you'd rather have those translated too, say so and I'll adjust — I just didn't want to
silently degrade generation quality or teach users an unparseable script format.

---

## Alternatives & context

Three things I ran into while doing this that may be worth separate attention.

> Everything below is read off the **built desktop bundles** of v2.0.0, not your source —
> I don't have the repo checked out. So variable names in the snippets are minified
> (`_e`, `rt`, `sB`) and line references are approximate. The behaviour is what I actually
> observed running the app.

### 1. Bug: the visual-style dropdown shows Chinese labels in English mode

This is not a Vietnamese problem — it affects existing English users today.

`GET /api/v1/styles` returns both a Chinese `label` and an English `name`:

```json
{ "id": "anime", "name": "Anime Style", "label": "动漫风格", "type": "preset" }
```

The ingest page builds its option list preferring `label`:

```js
_e.length > 0
  ? _e.map(rt => ({ value: rt.id, label: rt.label || rt.name || rt.id }))   // ← Chinese wins
  : sB.map(rt => ({ value: rt.value, label: t(rt.labelKey) }))              // ← i18n, never reached
```

Since the API always returns data, the `labelKey` branch never runs, so the dropdown reads
`动漫风格 / 写实古装剧 / 3D玄幻国漫 …` even with the UI set to English. The Styles page does
this correctly — it maps preset ids through the `ingest.visualStyles.*` key table first and
only falls back to the API label for user-created styles. Making the ingest dropdown use the
same resolution order fixes it for `en` and any future locale in one go.

### 2. A few places hard-branch on `zh` vs `en`, so a third locale degrades oddly

- The release-notifications locale resolver is `slice(0,2).toLowerCase() === "en" ? "en" : "zh"`,
  so any non-`en` locale fetches **Chinese** release notes. Inverting it (`=== "zh" ? "zh" : "en"`)
  makes unknown locales fall back to English, which is the friendlier default.
- `new Intl.RelativeTimeFormat(lang.startsWith("zh") ? "zh" : "en", …)` — `Intl` already
  supports `vi`, so passing the resolved language through gives correct relative dates for free.
- The account menu's "currently selected language" label is a binary
  `lang === "zh" ? languageChinese : languageEnglish`.

### 3. ~1273 Chinese strings are still hardcoded in the **frontend** bundles

#460 covers backend runtime text. The frontend has its own pile that never reaches i18next.
Counting distinct Chinese string literals in the v2.0.0 production bundles:

| chunk | distinct CJK literals |
|---|---|
| `freezone.lazy` | 880 |
| `index` | 240 |
| `characters.lazy` | 62 |
| `ThreeDDirectorDialog` | 53 |
| `beats.lazy` | 33 |
| others | 5 |
| **total** | **1273** |

Roughly **980 of those are genuinely user-visible labels** — Freezone skill-node definitions
(`label`, `description`, `shortName`, enum option labels), the 360 panorama viewer panel,
the new style-wall UI, Director World status/error messages, the landing page and FAQ copy,
and the asset breadcrumb templates. The other ~290 are prompt content, data values and
parser aliases that should stay Chinese.

**Not all of them should be extracted, though**, and the distinction is easy to get wrong:

- In `{ key:"shot_type", label:"景别", options:[{ value:"特写", label:"特写" }] }` only `label`
  is a UI string. `value` / `defaultValue` are data — the option-matching and the payload
  depend on them.
- `promptFragment: "镜头前推"`, `defaultOverride`, `"图片风格与参考图保持一致"`, `"禁止添加描述文本"`
  and the grid-prompt templates are prompt content.
- The storyboard-table column aliases (`镜号`, `画面描述`, `关键帧地址`, …), the script section
  labels matched by regex (`解说词`, `画面描述`, `置景描述`, `对白`, `出场人物`), the time-of-day
  values checked with `Set.has()` (`清晨`, `黄昏`, `夜晚`), and `"图"` in the `@图N` mention parser
  are all matched against user input and must stay as-is.

I'm not asking anyone to do this work — just flagging the shape of it, since it's the part
that can't be fixed by supplying a translation file, and getting the label/value split wrong
would break generation rather than just look untranslated.

### Note on the screenplay format guide

The format guide (`ingest.novelFormat.rule*`, `ingest.sceneHeaders.*`,
`aiAssistant.formatCheck.recommended`) documents the accepted formats. The spec block in the
code lists Chinese production formats and Fountain / Final Draft — no other keyword set. So a
localized guide has to describe the rules in the target language while keeping
`Episode N` / `第N集`, `Characters:` / `人物：`, `INT.`/`EXT.`, `DAY`/`NIGHT` literal, or users
follow the translated guide and their import fails. My `vi` file does it that way. Worth a
comment in the source if more locales land, because it's an easy trap.

---

Happy to open the PR whenever it's useful — or to hold off if you'd rather land `vi` after
the #460 extraction settles. Either way the file is ready.

hoanglongggggggggg:
Closes #489.

Adds the `vi` locale you green-lit, plus the two zh/en binary branches you asked me to fold in and the source comment about the format keywords. One extra fix came out of writing the locale — details in section 4.

Three commits, each independent and droppable.

---

## 1. `feat(i18n): add Vietnamese (vi) locale`

- `frontend/public/locales/vi/translation.json` — **5888 / 5888 keys**, exact parity with `en`
- `"vi"` in `SUPPORTED`
- `header.account.languageVietnamese` in `en` and `zh`
- parity test extended to every locale, not just `zh` vs `en`

Two small refactors came with it:

**The account menu no longer hardcodes two rows.** It maps over `SUPPORTED` with a `LANGUAGE_LABEL_KEYS` lookup, and `activeLanguage` goes through `normalize()` instead of `startsWith("zh") ? "zh" : "en"` (the third of the three binary branches from #489). Adding a fourth locale is now a tag plus a label key — no JSX.

**`SUPPORTED` / `normalize` moved to `i18n/languages.ts`.** `i18n/index.ts` initializes i18next as an import side effect, so importing it from `header.tsx` broke `header.test.tsx`, which mocks `react-i18next` without `initReactI18next`. A component that only needs the language list should not pull in the runtime. `index.ts` re-exports both, so existing imports are unaffected.

### What is deliberately left in English — 329 keys

Same reasoning as in #489, updated for what v2.0.2's extraction moved into the locale files:

| Group | Why |
|---|---|
| `nodeToolbar.*Prompt`, `lightEditor.prompt*` / `presetPrompts`, `node.imageEdit.suggest.*Prompt`, `episode.workbench.video.seedance2Guidance*Template` | Pure prompt content — composed into what goes to the image/video model, never rendered. Translating them measurably hurts output quality. The last group is newly extracted; `video-pane.tsx` says the guidance follows the UI language, so tell me if you'd rather have it in Vietnamese and I'll switch it. |
| `lightEditor.directions.*`, `multiAngleEditor.presets` / `zoomLevels` / `horizontalLabels` / `verticalLabels` | **These are dual-purpose and it's a real tradeoff, not a free win.** They render as visible labels (`LightEditorPanel.tsx:577`, `MultiAngleEditorPanel.tsx:228,258`) *and* get interpolated into the generated prompt (`LightEditorPanel.tsx:957`, `MultiAngleEditorPanel.tsx:515`). I chose prompt fidelity over a translated label, so a Vietnamese user sees "front" / "left" on those two panels. If you'd rather have the label translated, the fix is to split the key into a label and a prompt token rather than translating the shared one — happy to do that instead. |
| Domain terms — Model, Prompt, Render, Timeline, Keyframe, Upscale, Seed, Commit, canvas, node, asset, storyboard, beat | Vietnamese video/AI people use the English words; translating them reads worse than leaving them. |
| Product and proper nouns — SuperTale, Freezone, Director World, Shrimp Lens / Canvas / Pond / Feed, PlayCanvas, LingShan-G2, Grok Video Channel, HappyHorse | Names. |
| Slot and protocol identifiers quoted in UI copy — `scene_id`, `prop_id`, `identity_id`, `scene_master`, `prop_ref`, `selected_background`, `3GS`, `.sog`, GLB | They appear in the UI as the identifier the user has to type or match. |
| Screenplay-format keywords inside `ingest.novelFormat.*`, `ingest.formatCheck.*`, `aiAssistant.formatCheck.recommended` | See section 4. |

A few keys are identical to `en` simply because the Vietnamese happens to match (`Video`, `Album`, `FPV`, `Drone`, `Request ID`, the `", "` list separator). Those are in the 329 too.

---

## 2. `fix(i18n): fall back to English, not Chinese, for unpublished locales`

The other two binary branches from #489:

- `normalizeReleaseLocale` was `two === "en" ? "en" : "zh"` — every locale except literally `en` fetched **Chinese** release notes. Now only `zh` asks for Chinese.
- `formatRelativeTime` passed `startsWith("zh") ? "zh" : "en"` to `Intl`. The resolved language now goes straight through. Guarded with try/catch: `Intl.RelativeTimeFormat` throws on a structurally invalid tag, and a bad tag should not take the notification list down.

The existing test asserted the old behaviour (`defaults missing and unknown locales to zh`); it now asserts `en` and covers `vi-VN`.

---

## 3. Source comment about the format keywords

Added next to the format spec in `NovelFormatDialog.tsx`, naming the three key groups a translator will meet and stating that the parser matches the keywords literally. Also noted that the sample picker's `zh`-vs-`en` branch is binary **on purpose** — a locale without its own sample set should get the Fountain one — so a future reader doesn't "fix" it.

For the record, `Tập 1` and `Nhân vật: Ji-won` are what a Vietnamese writer would naturally type. The first fails to match an episode header. The second is worse: `parse_character_line` returns `["Nhân vật: Ji-won"]` — one character whose name is the whole line. It fails silently.

---

## 4. `fix(ingest): correct English screenplay-format guidance the parser rejects`

Not part of the original ask. I hit it while writing the Vietnamese format guide, because I copied your English one.

Three `en` strings teach a heading shape the parser does not accept. They were translated from the Chinese strings word for word, and `1-1 <location> <time> <int/ext>` only parses when the time and interior/exterior tokens are *Chinese*. Verified against `parse_scene_blocks` on current `main`:

```
EPISODE 1 / 1-1 City Cafe DAY INT     -> 1 block, location='' time='' int_ext=''
EPISODE 1 / INT. City Cafe - DAY      -> 1 block, location='City Cafe' time='日' int_ext='内'
EPISODE 1 / 1-1 苏鸾寝殿 深夜 内         -> 1 block, parses correctly
```

Affected:

- `aiAssistant.formatCheck.recommended` — the "standard script example" the assistant shows
- `ingest.formatCheck.issue.missingSceneHeaders.fix` — same example, in the fix hint
- `ingest.formatCheck.issue.sceneHeadersMissingTime.fix` — suggests `LATE NIGHT`, but `ENGLISH_TIME_TOKENS` is DAY / NIGHT / MORNING / AFTERNOON / EVENING / DAWN / DUSK. `INT. X - LATE NIGHT` does not match `INTERNATIONAL_SCENE_RE`, so the whole heading is lost — which `ingest.novelFormat.ruleTimeTokens` already warns about two screens away.

English-only; `zh` is correct as it stands. This is the trap from #489 turning out to be live rather than hypothetical, which is partly why the source comment in section 3 seems worth having.

---

## Verification

Everything CI runs, run locally on this branch:

| Check | Result |
|---|---|
| `scripts/check_frontend_i18n.py` | pass — 0 hardcoded Chinese, ratchet not moved |
| `scripts/lint_banned_words.py` | pass |
| `scripts/lint_ee_terms.py` | pass |
| `scripts/ce_allowlist.py` | pass |
| `scripts/check_dco.py` | pass — all 3 commits signed off |
| `pnpm build` (tsc -b + vite) | pass, `dist/locales/vi/translation.json` ships |
| `pnpm test` | 411 files, 2941 passed, **7 failed** |

Those 7 are pre-existing. On `upstream/main` with no changes the same run gives 2939 passed / **7 failed, identical tests**: 3 in `video-node-credit-contract`, 1 in `script-node-credit-contract`, 1 in `video-model-capabilities`, 2 in `auth-gating`. The auth-gating pair passes when its file runs alone, on both main and this branch — looks like a timeout under full-suite parallelism, not a real failure. My +2 passing are the extra parity case and the `vi-VN` assertion.

Locale integrity was checked mechanically, not by eye: key sets and key *order* identical to `en` across all three files; `{{...}}` / `<n>` / `$t()` placeholder multisets identical per key; no stray single braces; no key existing only as `_one` (Vietnamese resolves to the `other` category alone under `Intl.PluralRules`); and the `en`/`zh` diffs contain exactly one added key plus three changed `en` values, with every other value byte-identical to `upstream/main`.

Parser claims were checked by importing `novelvideo.utils.screenplay_scene_parser` and running `parse_scene_blocks` / `parse_character_line` / `ENGLISH_TIME_TOKENS` against the exact strings quoted above.

One convention worth stating so nobody "fixes" it piecemeal: the file uses old-style Vietnamese tone placement throughout — `hoà`, `khoá`, `xoá`, `tuỳ`, `huỷ`, `hoá` (190 occurrences, 0 of the modern `hòa` / `khóa` / `xóa` / `tùy` / `hủy`). Both are correct Vietnamese; this one is applied consistently. Happy to convert the whole file to the modern placement if you'd rather — it's one pass, but it should be all or nothing.

Terminology is likewise consistent rather than per-string: `cú máy` for shot, `khung đầu` for first frame, `bối cảnh` for scene-as-asset vs `cảnh` for scene-in-story, `ngữ cảnh cú máy` for shot context, `ứng viên` for candidate, `tạo hình` for identity, `phác thảo` for sketch, `luồng chính` for mainline, `dây chuyền` for pipeline.

@lywaterman

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -721,7 +721,7 @@
       "title": "Script Format Check",
       "close": "Close",
       "viewDetails": "View details",
-      "recommended": "Standard script example:\nEpisode 1\n1-1 City Cafe DAY INT\nCharacters: Lin Xia, Chen Mo\n△ Afternoon light crosses the window as Lin Xia places a letter on the table.\nLin Xia: The answer you wanted is right here.",
+      "recommended": "Standard script example:\nEPISODE 1\nINT. City Cafe - DAY\nCharacters: Lin Xia, Chen Mo\n△ Afternoon light crosses the window as Lin Xia places a letter on the table.\nLin Xia: The answer you wanted is right here.",
       "lineLabel": "Line {{line}}",
       "noIssues": "No specific issues found"
     },
@@ -2544,7 +2544,7 @@
         },
         "missingSceneHeaders": {
           "message": "No scene headings were recognized, so the system cannot tell where each scene begins.",
-          "fix": "Follow the standard format: start every episode with “Episode N”, give each scene its own “scene number, location, time, INT/EXT” heading, then list characters, action, and dialogue. For example: Episode 1 / 1-1 City Cafe DAY INT."
+          "fix": "Follow the standard format: start every episode with “EPISODE N”, give each scene its own “INT./EXT. location - TIME” heading, then list characters, action, and dialogue. For example: EPISODE 1 / INT. City Cafe - DAY."
         },
         "nonstandardSceneHeaders": {
           "message": "Scene boundaries were recognized, but some scene metadata is incomplete.",
@@ -2552,7 +2552,7 @@
         },
         "sceneHeadersMissingTime": {
           "message": "Some scene headings have no explicit time anchor, so time_of_day and scene variant inheritance may be unstable.",
-          "fix": "Add an explicit time to the scene heading, such as DAY, NIGHT, or LATE NIGHT. Scene boundaries are kept either way and normalized later."
+          "fix": "Add an explicit time to the scene heading. Only DAY, NIGHT, MORNING, AFTERNOON, EVENING, DAWN and DUSK are recognized. Scene boundaries are kept either way and normalized later."
         },
         "missingInteriorExterior": {
           "message": "The scene heading has no INT/EXT marker.",
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -436,6 +436,7 @@
       "selectLanguage": "Select language",
       "languageChinese": "中文",
       "languageEnglish": "English",
+      "languageVietnamese": "Tiếng Việt",
       "passwordDialog": {
         "title": "Change password",
         "description": "Changing your password signs you out on every device. Sign in again with the new password.",
--- a/frontend/public/locales/vi/translation.json
+++ b/frontend/public/locales/vi/translation.json
@@ -1,0 +1,7414 @@
+{
+  "app": {
+    "title": "SuperTale",
+    "subtitle": "Bàn làm việc tạo và commit asset của SuperTale",
+    "updateRequired": {
+      "title": "App đã được cập nhật",
+      "description": "Trang này đang dùng file app đã hết hạn. Tải lại để dùng bản mới nhất.",
+      "refresh": "Tải lại trang"
+    },
+    "updateAvailable": {
+      "title": "Đã có phiên bản mới",
+      "description": "Tải lại trang để nhận các tính năng và bản sửa lỗi mới nhất.",
+      "refresh": "Tải lại ngay",
+      "dismiss": "Bỏ qua"
+    },
+    "routeError": {
+      "title": "Không tải được trang",
+      "description": "Hãy tải lại trang và thử lại."
+    },
+    "versionUpdate": {
+      "title": "Tính năng mới đã có mặt",
+      "confirm": "Đã hiểu",
+      "empty": "Phiên bản này không có ghi chú phát hành"
+    },
+    "logoHomeTooltip": "Bấm để quay về trung tâm quản lý dự án"
+  },
+  "loginCinematic": {
+    "moreInfo": "Xem thêm thông tin",
+    "moreInfoShort": "Thêm",
+    "announcement": {
+      "open": "Xem thông báo",
+      "label": "Thông báo",
+      "title": "Thông báo",
+      "close": "Đóng thông báo",
+      "confirm": "Xác nhận",
+      "markAllRead": "Đánh dấu đã đọc tất cả",
+      "unread": "{{n}} chưa đọc",
+      "pinned": "Đã ghim"
+    },
+    "testimonials": {
+      "heading": "Không còn trang giấy trắng",
+      "subheading": "Phim ngắn concept, thử nhân vật và đoạn phim liền mạch — đi từ một câu tiền đề thẳng vào dây chuyền.",
+      "quotes": {
+        "shortFilmDirector": {
+          "name": "Đạo diễn phim ngắn",
+          "tag": "Previz concept",
+          "text": "Không còn phải bắt đầu từ một timeline trắng. Thấy xung đột và cú máy trước, rồi mới quyết nhánh nào đáng làm."
+        },
+        "aiCreator": {
+          "name": "Người làm video AI",
+          "tag": "Đoạn phim liền mạch",
+          "text": "Cái đáng giá không phải tạo ra một tấm ảnh — mà là đẩy được một đoạn phim đi tiếp."
+        },
+        "screenwriter": {
+          "name": "Biên kịch",
+          "tag": "Thử nhân vật",
+          "text": "Khi nhân vật, xung đột và bối cảnh đã tách ra, tôi nhận ra nhanh hơn hẳn một câu chuyện có đáng viết hay không."
+        },
+        "animationTeam": {
+          "name": "Nhóm hoạt hình",
+          "tag": "Kiểm tra nhịp",
+          "text": "Bọn tôi dựng previz concept trước — soát nhịp và cú máy — rồi mới quyết có đưa vào sản xuất không."
+        },
+        "indieProducer": {
+          "name": "Nhà sản xuất độc lập",
+          "tag": "Dựng trailer",
+          "text": "Nó nén một ý tưởng thành thứ xem được, nên cuộc bàn bạc không còn dừng ở mấy dòng ghi chú."
+        },
+        "visualDirector": {
+          "name": "Giám đốc hình ảnh",
+          "tag": "Mở rộng thế giới",
+          "text": "Một thế giới cứ mọc thêm nhánh mới. Vẫn hoang dại, nhưng hướng đi thì vẫn lái được."
+        },
+        "storyPlanner": {
+          "name": "Người lên nội dung",
+          "tag": "Chọn nhánh",
+          "text": "Tôi dùng nó để loại nhanh những nhánh cụt và dồn thời gian cho nhánh thật sự có sức căng."
+        },
+        "creativeStudio": {
+          "name": "Studio sáng tạo",
+          "tag": "Bản mẫu chào hàng",
+          "text": "Từ một câu tiền đề tới một đoạn phim xem được — đủ để dẫn dắt một cuộc bàn sáng tạo cụ thể hơn nhiều."
+        },
+        "directorAssistant": {
+          "name": "Trợ lý đạo diễn",
+          "tag": "Gom cú máy",
+          "text": "Cú máy thôi tản mát. Mỗi lượt tạo đều rơi về một đường ray đẩy tiếp được."
+        }
+      }
+    },
+    "faq": {
+      "heading": "Hỏi gì đáp nấy, không vòng vo",
+      "subheading": "Về khâu tạo nội dung, khả năng kiểm soát, làm việc nhóm và hợp tác kinh doanh — chỉ những câu thật sự làm bạn đổi quyết định.",
+      "contactPrompt": "Có câu hỏi hợp tác cụ thể?",
+      "contactOpen": "Mở liên hệ kinh doanh",
+      "contactCta": "Liên hệ kinh doanh",
+      "contactLabel": "Liên hệ kinh doanh",
+      "qrAlt": "Mã QR WeChat doanh nghiệp",
+      "items": {
+        "difference": {
+          "question": "DramaClaw khác gì một công cụ tạo video AI thông thường?",
+          "answer": "Công cụ thông thường xoay quanh một prompt, một đoạn clip hay một tấm ảnh. DramaClaw dựng quanh cả một dự án hoàn chỉnh: nhập chữ, gây dựng asset, lên kế hoạch tập, sản xuất cú máy và dựng bản cuối — một vòng sản xuất truy vết được, làm nhóm được, dùng lại được."
+        },
+        "gettingStarted": {
+          "question": "Lần đầu dùng thì nên bắt đầu từ đâu?",
+          "answer": "Cứ đi theo luồng chính: nhập văn bản ở Shrimp Feed, xác nhận nhân vật / bối cảnh / đạo cụ / giọng ở Shrimp Pond, lên kế hoạch tập rồi tạo kịch bản và cú máy ở Shrimp Lens, cuối cùng xuất file dựng ở trang dựng phim. Mục tiêu đầu tiên không phải là hoàn hảo — mà là đưa trọn một tập đi hết dây chuyền."
+        },
+        "mirrorVsCanvas": {
+          "question": "Shrimp Lens và Shrimp Canvas mỗi cái mạnh ở đâu?",
+          "answer": "Shrimp Lens là dây chuyền sản xuất chính: làm hàng loạt, ổn định, tiến từng tập một. Shrimp Canvas là bàn làm việc của đạo diễn: chau chuốt những cú máy then chốt, thử nhiều phiên bản, chốt asset và xử lý phần video nặng. Trong một dự án thật, những cú máy thông thường chạy qua Shrimp Lens còn cú máy khó thì đưa sang Shrimp Canvas."
+        },
+        "canvasOverwrite": {
+          "question": "Kết quả trong Shrimp Canvas có tự ghi đè asset luồng chính của tôi không?",
+          "answer": "Không. Mọi thứ tạo ra hay tải lên trong Shrimp Canvas mặc định chỉ là bản ứng viên. Nó chỉ vào luồng chính của dự án khi bạn chủ động commit và chọn slot đích — nhân vật, bối cảnh, đạo cụ, phác thảo beat, khung đầu hay video."
+        },
+        "whyAssetsFirst": {
+          "question": "Vì sao phải sắp xếp nhân vật, bối cảnh, đạo cụ và giọng trước?",
+          "answer": "Nguyên nhân thường gặp khiến phải làm đi làm lại trong video AI không nằm ở khâu tạo cuối cùng — mà ở phía trên: nhân vật không ổn định, bối cảnh mơ hồ, đạo cụ không có ảnh mẫu, giọng không nhất quán. Biến những thứ đó thành asset ở Shrimp Pond sẽ gỡ được rất nhiều rối rắm ở khâu cú máy."
+        },
+        "contentTypes": {
+          "question": "DramaClaw hợp với những loại nội dung nào?",
+          "answer": "Phim ngắn AI, phim truyện tranh, trailer tiểu thuyết, video bình phim, quảng cáo và video đào tạo — cùng với những dự án nhóm cần duy trì nhân vật IP và nội dung ra theo tập trong thời gian dài. Người làm một mình chạy được cả dây chuyền; nhóm thì sản xuất cùng nhau được."
+        },
+        "teamRoles": {
+          "question": "Một nhóm nên chia việc thế nào?",
+          "answer": "Chia theo công đoạn: chủ dự án lo quản lý dự án, chi phí và kiểm định; người viết lo văn bản và kịch bản; chỉ đạo mỹ thuật giữ nhân vật, bối cảnh và đạo cụ; người làm storyboard/đạo diễn lo diễn xuất trong từng cú máy; người dựng video lo khung đầu, video và ghép phim. Quyền hạn tương ứng: người xem, người sửa, quản trị và chủ sở hữu."
+        },
+        "unstableResults": {
+          "question": "Khi kết quả lúc được lúc không thì nên kiểm tra gì trước?",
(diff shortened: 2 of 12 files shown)


n7s4 (commented):
## Review — reviewed locally at head `f2e88ecd` against base `main@a7d48af3`

Reviewed in a dedicated local worktree (`git fetch refs/pull/500/head`), not via the GitHub file view. The locale file was checked mechanically rather than by eye; results are in the verification table at the end.

**Summary:** the engineering is solid and the mechanical integrity of the locale is clean — key parity, key order, placeholder multisets, newline counts, plural categories and "no Chinese leaked into `vi`" all pass. **Nothing here is a merge blocker:** no crash, no data corruption, no permission/migration/compatibility risk. Everything below is copy correctness in the screenplay-format guidance — where a wrong string sends a writer down a path the parser rejects, but the format check does surface that and blocks the import rather than corrupting anything.

*(Corrected: an earlier version of this review graded the first finding P0. That was miscalibrated — it is a wrong sentence in a help dialog, recoverable, with an in-product error pointing at it. Re-graded below.)*

---

### P1 — `vi` `ruleLocationChange` was translated from `zh`, and teaches a heading order the parser rejects under English keywords

`frontend/public/locales/vi/translation.json:2467` (rendered at `frontend/src/components/ingest/NovelFormatDialog.tsx:246`)

```
zh: "地点 时间 内/外"和"地点 内/外 时间"两种顺序都支持。
en: Open a new scene heading whenever the physical location changes. Use INT. for interior
    and EXT. for exterior — a complete heading needs one of the two.
vi: Tạo dòng đầu cảnh mới mỗi khi đổi địa điểm. Hỗ trợ cả thứ tự Địa điểm-Thời gian-INT/EXT
    lẫn Địa điểm-INT/EXT-Thời gian.
```

This one followed `zh` instead of the `en` you corrected in section 4. Both of those orders are accepted **only** by the Chinese branch of `parse_scene_blocks` — they need Chinese time tokens (`深夜`, `日`, …) and `内` / `外`. Everywhere else in the same dialog the Vietnamese text correctly instructs the reader to keep the English keywords (`ruleScene`, `ruleTimeTokens`, `ruleEpisode` all say *"giữ nguyên"*), and with English keywords only `INTERNATIONAL_SCENE_RE` (`INT. <location> - <TIME>`) matches.

Failure: a Vietnamese user follows this rule, writes `City Cafe DAY INT`, and the heading is not recognized. The format check does catch it (`Không nhận ra dòng đầu cảnh nào`, blocking status), so this is a wrong-guidance bug rather than a silent corruption — but it is guidance shipped by this PR, in the feature this PR exists to add, and `ruleScene` two lines above already gives the correct primary shape. The `en` sentence "a complete heading needs one of the two" is also dropped entirely in `vi`.

P1 rather than a blocker: recoverable, surfaced in-product, one line to fix. Suggested: translate from the corrected `en` string, not from `zh`.

---

### P2 — the section-4 fix is incomplete: two more `en` strings still teach the shape the parser rejects

Same defect class, same file, small diff. Both are pre-existing on `main`, so strictly out of scope — raised because this PR is the one that went looking for them, and because `vi` inherits one and contradicts the other.

**a) `frontend/public/locales/en/translation.json:2474` — `ingest.sceneHeaders.missingFix`**

> …then give every scene a separate **"scene number, location, time, INT/EXT"** heading…

Rendered at `frontend/src/routes/_app/projects.$project/ingest.tsx:329`. This is the same `1-1 <location> <time> <int/ext>` shape you removed from `aiAssistant.formatCheck.recommended` and `ingest.formatCheck.issue.missingSceneHeaders.fix`. `vi:2474` carries it over faithfully.

**b) `frontend/public/locales/en/translation.json:2560` — `ingest.formatCheck.issue.missingInteriorExterior.fix`**

> Add INT or EXT, as in **"location time INT/EXT"**.

This is live UI copy: the backend emits `missing_interior_exterior` (`src/novelvideo/utils/screenplay_quality.py:501,542`) and it is resolved through `frontend/src/lib/format-check-copy.ts:93`. Also Chinese-only order.

Note that `vi:2560` **fixes** this one (`dạng "INT. địa điểm - DAY"`), so after merge `en` and `vi` give contradictory advice for the same key. Either fix `en` too, or say why only half.

---

### P2 — `vi` `ruleDialogue` drops the parser trap the `en` string exists to warn about

`frontend/public/locales/vi/translation.json:2466`

`en` says explicitly: *"a bare uppercase character name on its own line is not recognized as a speaker."* The Vietnamese version keeps only the shape (`"TênNhânVật: lời thoại"`) and loses the warning. By the standard you set in section 3 — the keyword traps are the part a translator must preserve — this one should come back.

---

### P2 — login-page announcements still fall back to Chinese before English

`frontend/src/components/login/cinematic/announcements.ts:138`

```ts
const candidates = [tag, tag.split("-")[0], "zh", "en"];
```

A Vietnamese user gets **Chinese** announcements on the login page, while commit 2 just moved release notes the other way (`release-notifications.ts:41`). Same family of binary branch from #489, and the reasoning in your own comment ("far more likely to be a reader of English than of Chinese") applies identically here. Either swap to `"en", "zh"` or note why the two surfaces should differ.

---

### P3 — the new parity invariant isn't reflected in the contributor-facing hints or the other locale lists

The extended test in `frontend/src/__tests__/i18n/locales-json.test.ts` now makes `vi` a hard requirement for every new key. Three places still assume two locales:

- `scripts/check_frontend_i18n.py:174` — *"…在 frontend/public/locales/{zh,en}/translation.json 补齐两边的 key"*
- `scripts/check_backend_i18n.py:178` — same wording

  A contributor who follows either hint adds the key to `zh`/`en` only and gets a red parity test with no clue that `vi` was the missing half.

- `frontend/src/__tests__/features/canvas/capability-i18n.test.ts:24` and `frontend/src/__tests__/lib/model-task-access.test.tsx` still hardcode `["zh", "en"]` rather than reading `SUPPORTED`. Not a bug (parity covers key existence), but the locale list is now duplicated in four places and only one of them moved.

### P3 — smaller things

- **`frontend/src/__tests__/components/layout/header.test.tsx:66-68`** — the `t` mock doesn't define `header.account.languageVietnamese`, and falls through to returning the key. The new third row renders as a raw key string in the test DOM and nothing asserts on it, so the row the refactor added is effectively uncovered.

- **`frontend/public/locales/vi/translation.json:5087` — `toolDialog.suffix` is `""`** (`en: " Tool"`, `zh: "工具"`). It's the only leading/trailing-whitespace divergence in the whole file, and it reads as a deliberate choice, since a Vietnamese qualifier would go before the noun, not after. Worth a comment though: it only renders as "no suffix" because i18next's `returnEmptyString` defaults to `true` and `i18n/index.ts` doesn't override it. If anyone ever sets it to `false`, this silently becomes the Chinese "工具" via `fallbackLng: "zh"`.

- **`frontend/src/features/canvas/pricing/index.ts:70,92,215`** — a `vi` user now gets `en-US` number formatting, USD, and a hardcoded English `/run` suffix. Pre-existing and out of scope, but the `// i18n-exempt —— 已按语言分支` marker on line 215 rests on a premise ("the branch covers the languages") that this PR ends. Worth a follow-up issue rather than a change here.

- **PR description drift.** The body says "three commits, each independent and droppable"; the branch is four commits plus `da3e9be3 Merge branch 'main'`, and the last one (`f2e88ecd fix(compliance): register Vietnamese locale files`) is authored by a different contributor. The verification table therefore predates the current head by two commits. Worth refreshing the body, or squashing, before merge.

---

### Verification I ran

Mechanical checks against head `f2e88ecd`, all on the locale files themselves:

| Check | Result |
|---|---|
| Leaf key count en / zh / vi | 5892 / 5892 / 5892 — 0 missing, 0 extra |
| Key **order** identical to `en` | yes |
| Placeholder multisets (`{{…}}`, `<n>`, `$t()`) per key | 0 mismatches |
| Stray single braces | 0 |
| Newline count per key | 0 mismatches |
| Leading/trailing whitespace divergence | 1 (`toolDialog.suffix`, above) |
| `vi` values containing CJK | 10 — all reviewed, all legitimate (quoted prompt literals, `中文` / `Tiếng Việt` language names, the author name `痕继痕迹`). No leaked Chinese UI copy. |
| `vi` value == `zh` value where `en` != `zh` | **0** |
| `vi` value == `en` value | 328 — walked namespace by namespace, all fall inside the groups declared in the PR body |
| Plural bases (5) | every one has `_other`; safe under Vietnamese's single plural category |
| Type mismatches / unexpected empties | 0 / 1 (the `suffix` above) |
| Parser claims in section 4 | confirmed by reading `screenplay_scene_parser.py`: `ENGLISH_TIME_TOKENS` is exactly the seven tokens and has no `LATE NIGHT`; `INT. City Cafe - DAY` matches `INTERNATIONAL_SCENE_RE`; `Characters:` is accepted by `LABELED_CHARACTER_RE` |
| DCO | all four non-merge commits signed off; the merge is skipped by `scripts/check_dco.py` as designed |
| REUSE / license inventory | `REUSE.toml` is a `**` aggregate and `frontend/REUSE.toml` has `public/locales/**`, so only `license-inventory.csv` needed the two new rows — both present and correctly sorted |
| `lint_banned_words` / `lint_ee_terms` scope | neither scans `.json` or `.ts`, so the new files are out of their reach |

Three verification lanes, kept separate:

1. **Local test run on the PR head** — **not performed.** Running the branch's test/build scripts executes code from the PR, so I left that to you; everything above is static analysis plus data checks on the JSON.
2. **Merge against current base** — `git merge-tree --write-tree` against `main@a7d48af3` exits 0, no conflicts. (The branch already merged that same commit in `da3e9be3`, so it is fully current.)
3. **GitHub CI** — `gh pr checks` reports *no checks* on this branch. The verification table in the PR body is the only test signal, and it predates the current head.

To be explicit about the grading: **no P0.** The single P1 is one sentence in `vi:2467`; everything else is P2/P3 and several items are pre-existing rather than introduced here.

n7s4 (approved):

lywaterman:
Merged, and thank you. This is the first community locale in DramaClaw and it set the bar high: a complete, verified translation, the three `zh`/`en` binary branches fixed so any future locale falls back to English, tests for the release-notes resolver, the license inventory kept in sync, and a clear account of what was deliberately left in English and why. One review round, nothing to push back on.

Vietnamese will ship in the next release. If you have a moment, the README translation in `readme/README_vi.md` could use a native speaker's pass, no pressure. And if you'd like to keep `vi` current going forward, we'd be glad to have you as its maintainer; just say the word.

Cảm ơn bạn.

PR #528, opened 2026-09-10
fix: support reasoning-only Recipe compiler

vkiki:
Recipe compilation failed when the upstream model rejected `reasoning_effort=none` because the selected model always uses reasoning.

This change defaults Recipe compilation to `low`, allows `low/high/max` through `FREEZONE_RECIPE_COMPILER_REASONING_EFFORT`, and rejects unsupported values locally before the upstream request.

Validation:

- `PYTHONPATH=src /Users/tao/Works/SuperTale/.venv/bin/python -m pytest tests/test_recipe_runtime.py tests/test_project_lifecycle_storage.py -q` (`34 passed`)
- Ruff passed for the remaining changed Python files
- `git diff --check`


vkiki:
已按最新排查结论撤回本 PR 的永久删除与 Recipe reasoning 配置改动；先保留现有通用行为，待取得环境证据后再分别处理。

PR #585, opened 2026-09-16
fix: forward PYTHONPATH to the DramaClaw MCP servers

bopy-zou:
## Why

After pinning `mcp` (#584), Codex sessions still fail on deployments where `novelvideo` is installed into site-packages and also made importable from a source tree via `PYTHONPATH` (the API process resolves the source tree first):

```
required MCP servers failed to initialize: dramaclaw: Mcp error: 0: [Errno 2] No such file or directory: '/usr/local/lib/python3.12/.hermes/plugins/dramaclaw/__init__.py'
```

Codex starts stdio MCP servers with a minimal environment and forwards only the names listed in `env_vars`. `PYTHONPATH` was not listed, so the child `python -m novelvideo.chat.dramaclaw_mcp` imported the site-packages copy, and `_repo_root()` (`Path(__file__).resolve().parents[3]`) resolved to the Python prefix instead of the checkout root where `.hermes/plugins/dramaclaw` lives. Reproduced by running the server with and without `PYTHONPATH`: `tools/list` fails with the Errno 2 above without it and succeeds with it.

## What

Add `PYTHONPATH` to `env_vars` of both the `dramaclaw` and `dramaclaw_workflows` stdio servers in `_dramaclaw_mcp_servers`. Forwarding is value-based (`backend_sdk` only copies listed variables that are set in the parent env), so deployments without `PYTHONPATH` are unaffected. Tests updated for the whitelist and the serialized overrides.

## Verification

`pytest tests/test_chat_service_user_agent_scope.py -k "mcp or codex"`: 143 passed. `lint_ee_terms.py` / `lint_banned_words.py` / ruff clean.


vkiki:
Superseded by #586. The runtime-root resolver and DRAMACLAW_ROOT forwarding in #586 cover the deployment failure, and commit 40140ce4 also carries the PYTHONPATH forwarding from this PR for the workflow MCP server.

